PR #1, opened 2020-10-02
Merge the inital set of examples.

alex4o:
These are most of the simple examples I could think of that show what capabilities are doing.


common.h
```diff
@@ -0,0 +1,30 @@
+#include<stdio.h>
+#include<stdint.h>
+//#include<cheri/cheri.h>
```
ltratt: Generally we don't want to include commented code.

common.h
```diff
@@ -0,0 +1,30 @@
...
+	uint64_t perms = cheri_perms_get(ptr);
+	uint64_t type = cheri_type_get(ptr);	
+	bool tag = cheri_tag_get(ptr);
+	bool valid = cheri_is_valid(ptr);
+
+	uint64_t offset = __builtin_cheri_offset_get(ptr);
+	printf("Address: %04lx, Base: %04lx, End: %04lx Flags: %04lx, Length: %04lx, Offset: %04lx, Perms: %04lx, Type: %04lx, Tag: %d, Valid: %d\n", address, base, base + length, flags, length, offset, perms, type, tag, valid);
+}
+
+// This hsould be a macro or inlined
```
ltratt: Why?

function.c
```diff
@@ -0,0 +1,31 @@
...
+    if (a == b) 
+        return a; 
+    if (a > b) 
+        return gcd(a-b, b); 
+    return gcd(a, b-a); 
+} 
+
+
+int main() {
+
```
ltratt, line 21: Extraneous blank lines at beginning/end of function.

hello.c
```diff
@@ -0,0 +1,27 @@
+#include "common.h"
+//#include <cheri/cheric.h>
+#include <cheriintrin.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+
+/***
+ * This is my scratch file for testing things
```
ltratt, line 10: Out of date comment?

mmap.c
```diff
@@ -0,0 +1,68 @@
...
+	code[idx++] = clc_128(cra, csp, 16); //0x0101208F; // clc cra, 16(csp)
+	code[idx++] = cincoffsetimm(csp, csp, 32); //0x0201115B; // cincoffset csp, csp, 32
+	code[idx++] = cjalr(zero, cra); //0xFEC0805B; // cret
+}
+
+int main()
+{
+	code = get_executable_block();
+	generate(code);
+
```
ltratt: Lots of whitespace!

mmap.c
```diff
@@ -0,0 +1,68 @@
...
+
+
+	// Very important, the flags of the pointer that is used to call a function are then transfered to the pcc register
```
ltratt, line 41: Can we wrap to a smaller column width? Some people prefer 80; I tend to go to 100. I don't mind too much, but we certainly shouldn't go beyond, say, 120.

mmap.c
```diff
@@ -0,0 +1,68 @@
...
+	// The flags in the pcc reigster matter when executing: https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/8253bffe30abf2a8ae1c7eba515061b141aff727/src/cheri_addr_checks.sail#L109
+	// The registyer is used to determine the mode of operation. Is it in capability mode or in hybrid mode.
+	// When in hybrid mode the capabilities of some registers don't matter and the ddc (default data capability) is being used.
+	// When in capability mode the capability of the coresponding capability register is used, if you try to access a0 that would use the capability inside of ca0
+	//
+	code = cheri_flags_set(code, 0x0001);
+	inspect_pointer(cheri_pcc_get());
+	inspect_pointer(code);
+
+
```
ltratt, line 50: Commented code.

alex4o:
It should be ready for review again.

mmap.c
```diff
@@ -54,10 +46,42 @@ int main()
 	// When in capability mode the capability of the coresponding capability register is used, if you try to access a0 that would use the capability inside of ca0
 	//
 	code = cheri_flags_set(code, 0x0001);
+
+}
+
+// The hybrid function works because the DDC can be set to the stack pointer and used as a referance 
+// This because when we are in hybrid mode the DDC's address is what is used, and here we have it set to the stack pointer.
+// This means that when we want to store a value on the stack we can just do `csc cra, 16(cnull)`.
+// The calculateion is dcone like this: `ddc.address + cnull_value + 16`.
```
ltratt: Couple of typos :)

stackscan.c
```diff
@@ -0,0 +1,142 @@
...
+}
+
+void scan_all() {
+	void *stack_bot = __builtin_frame_address(0);
+	scan_range(stack_top, stack_bot);
+}
+
+int test_2()
+{
+	//void *cs0;
```
ltratt: Not sure what the commented code is telling us?

ltratt:
OK, I've done a *very* shallow initial review: I now want to come back and look at things with a bit more of a "big picture" hat on. That might take me a day or so.

mmap.c
```diff
@@ -0,0 +1,102 @@
...
+
+/* 
+ * This hybrid function works because the DDC can be set to the stack pointer and used as a referance.
+* This works because when we are in hybrid mode the DDC's address is used as base.
+* We have the ddc set to be the stack pointer. 
+* This allows storing values on the stack though `csc cra, 16(zero)`.
+* The calculateion is done like this: `ddc.address + zero_value + 16`.
+* Therefore cnull can be used in place of the stack pointer in the case of this example.
+*/
+uint32_t *generate_hybrid(uint32_t *code)
```
cristianurlea, line 65: Ditto as per `generate_purecap` comment.

mmap.c
```diff
@@ -0,0 +1,102 @@
...
+	code[idx++] = cincoffset(cs0, csp, 32);
+	code[idx++] = addi(a0, zero, 5
+	code[idx++] = clc_128(cs0, zero, 0);
+	code[idx++] = clc_128(cra, zero, 16);
+	code[idx++] = cincoffsetimm(csp, csp, 32);
+	code[idx++] = cjalr(zero, cra);
+	return code;
+}
+
+uint32_t *generate_micro(uint32_t *code)
```
cristianurlea, line 81: Ditto as per `generate_purecap` comment.

set_bounds.c
```diff
@@ -0,0 +1,28 @@
...
+ * The point of this program is to demonstrate the failing of a program
+ * that tries to extend the bounderies of a capable pointer.
+ ***/
+int main()
+{
+	int32_t array[16] = {0};
+
+	uint32_t bounds = 64;
+	puts("bounds(64): ");
+	scanf("%u", &bounds);
```
cristianurlea, line 17: test `scanf` error code and short-circuit if needed.

stackscan.c
```diff
@@ -0,0 +1,147 @@
...
+
+bool is_stack_pointer(void *ptr)
+{
+	if (cheri_is_valid(ptr))
+	{
+		uint64_t base = cheri_base_get(stack_top);
+		uint64_t length = cheri_length_get(stack_top);
+		uint64_t address = cheri_address_get(ptr);
+
+		if (address >= base && address <= (base + length))
```
cristianurlea, line 24: ()++

stackscan.c
```diff
@@ -0,0 +1,147 @@
...
+	if (cheri_is_valid(ptr))
+	{
+		return true;
+	}
+	return false;
+}
+
+bool is_exec(void *ptr)
+{
+	if (((cheri_perms_get(ptr) & 0b10) >> 1) == 1)
```
cristianurlea, line 43: literal on LHS of equality check

stackscan.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+int test_3()
+{
+	return test_2();
+}
+
+uint32_t *test()
+{
+	uint32_t *values = (uint32_t *)malloc((size_t)test_3());
```
cristianurlea, line 125: Some spaces between atom names would be nice.

allocate.c
```diff
@@ -0,0 +1,12 @@
+#include "include/common.h"
+#include <stdio.h>
+#include <stdlib.h>
+
+int main()
+{
+	size_t size = 31;
```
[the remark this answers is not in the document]
ltratt: I don't think it needs initialising at all?

allocate.c
```diff
@@ -0,0 +1,12 @@
...
+	printf("Size: ");
+	scanf("%lu", &size);
```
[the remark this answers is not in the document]
ltratt: For a small example program, I think it's probably OK not to check.

function.c
```diff
@@ -0,0 +1,33 @@
+#include "include/common.h"
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int gcd(int a, int b)
```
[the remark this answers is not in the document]
ltratt: I think `const` is not necessary in general, and certainly not in a small example ;)

allocate.c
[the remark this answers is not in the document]
cristianurlea: If `scanf` falls through, say with a trivial typo of `] <enter-key>' , if one tried to "default answer" the input request, you'd get a random value for `size` and an odd allocation attempt. 

```
Size: 17401064535334696868
Address: 0000, Base: 0000, End: ffffffffffffffff Flags: 0000, Length: ffffffffffffffff, Offset: 0000, Perms: 0000, Type: ffffffffffffffff, Tag: 0, Valid: 0
```
If you did initialize this to `0` however,  an erroneous input of `] <enter-key>'  gives a valid allocation ( of nil size, as per `malloc` spec , such that you can subsequently call `free` ):

```
Size: 0
Address: 407a8010, Base: 407a8010, End: 407a8010 Flags: 0000, Length: 0000, Offset: 0000, Perms: 6817d, Type: ffffffffffffffff, Tag: 1, Valid: 1
```

If you leave it initialised with `31`, the same typo gives: 

```
Size: 31
Address: 407c1000, Base: 407c1000, End: 407c101f Flags: 0000, Length: 001f, Offset: 0000, Perms: 6817d, Type: ffffffffffffffff, Tag: 1, Valid: 1
```

allocate.c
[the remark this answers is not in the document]
cristianurlea: I get that we're not processing stock market trades, but I have been known to loose entire days debugging single character typos before :). 

As small as it may be, it  is also a perfect example of how errors compound in the land of C.  Either a safe initialization value for `size` or a branch based off the return code of `scanf` would save you from the typo scenario above. Both would countermeasures would better and give redundancy for future code edits.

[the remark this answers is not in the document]
ltratt: I think for the small example programs here, some shortcuts are acceptable: there is no pretence that this is production code. Of course, we don't want to let active *bugs* in!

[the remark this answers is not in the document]
ltratt: Though I do agree that `scanf` is a bit easy for a user to get wrong so something simple like:

```c
if (scanf(...) < 1)
    err(...);
```

is probably a reasonable compromise.

allocate.c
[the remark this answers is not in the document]
alex4o: The reason I left it to 31, was to have a default value in case there is nothing written.

common.h
cristianurlea: I'm only guessing  but the (currently commented out) use of `inspect_pointer` in the **stackscan.c** example indicates we may wish to use `inspect_pointer` whilst traversing and inspecting the stack. Inlining via macro expansion avoids the issue of mutating the stack if the call to `inspect_pointer` is not otherwise inlined by the compiler.

alex4o: Added an answer it the comment: https://github.com/capablevms/cheri-examples/blob/1f8537c5e8e99633fb816022662eb0d728c14d84/include/common.h#L28-L34

common.h
alex4o: Fixed in: https://github.com/capablevms/cheri-examples/pull/1/commits/1f8537c5e8e99633fb816022662eb0d728c14d84

hello.c
alex4o: Fixed in: https://github.com/capablevms/cheri-examples/pull/1/commits/1f8537c5e8e99633fb816022662eb0d728c14d84

mmap.c
alex4o: Formated in: https://github.com/capablevms/cheri-examples/pull/1/commits/1f8537c5e8e99633fb816022662eb0d728c14d84

mmap.c
alex4o: I rewrote this and, the column length now is under 100, in https://github.com/capablevms/cheri-examples/pull/1/commits/9923659a2770c9656d1afc8b7736dd3d3b7f504c

mmap.c
alex4o: Fixed in: https://github.com/capablevms/cheri-examples/pull/1/commits/9923659a2770c9656d1afc8b7736dd3d3b7f504c

function.c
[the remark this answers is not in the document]
alex4o: Full disclosure, I didn't write this function. I copied it from the internet, just to serve as a small example.

stackscan.c
alex4o: ???

allocate.c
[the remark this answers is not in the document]
alex4o: Fixed like that: https://github.com/capablevms/cheri-examples/blob/6332f5d4b12ba8ded2e0f687d4e52f748b6ae534/set_bounds.c#L17-L19

hello.c
```diff
@@ -0,0 +1,23 @@
...
+int main()
+{
+	int32_t array[16] = {0};
+
+	int64_t type = 4;
+	int32_t *typed_array = &array;
+	int32_t *aptr = &array;
+
+	uint64_t length = cheri_length_get(typed_array);
+	for (uint32_t counter = 0; counter <= length / sizeof(int32_t) + 16; counter++)
```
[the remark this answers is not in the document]
alex4o: Should be fine now: https://github.com/capablevms/cheri-examples/blob/0dee6e6f5890d70122c8837b0034d40b973b4dc1/hello.c#L17

mmap.c
```diff
@@ -0,0 +1,102 @@
...
+#include "include/regs.h"
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/cdefs.h>
+#include <sys/mman.h>
+
+uint32_t *get_executable_block()
```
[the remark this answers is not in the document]
alex4o: There is, you can always use `cheri_get_lenght` to obtain the size of the memory.

stackscan.c
alex4o: The comments were removed.

stackscan.c
alex4o: clang-format disagrees

alex4o:
I am once again asking for your reviews.

stackscan.c
cristianurlea: I would like to challenge `clang-format` to a rock paper scissors duel, 3/5 wins.
Default options are not always the best options.

stackscan.c
cristianurlea: I find explicit evaluation order more appealing:  `( (address >= base) && ( address <= (base + length)) )`. 
I would also guard `base`, `length` and `address` with `const` if the intent is to use them for a single bounds check before they go out of scope, but that may be a bit much for this example.

mmap.c
cristianurlea: Perhaps check-in `ColumnLimit: 100` in `.clang-format` too ?

common.h
cristianurlea: I think the reasoning is sound, roughly the same as the above comment regarding `inspect_pointer`.
Using the `inline` keyword will still require that you provide a non-inline implementation that manipulates the `csp` value.
Using a macro has the advantage that you're not dependent on the optimisation level selected during compilation.

Out of these two I would still prefer the `inline` keyword version. I'm happy to `resolve` the issue subject to an issue being opened describing the chosen course of action, unless there is any objection.

mmap.c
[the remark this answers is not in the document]
cristianurlea: That's the single most intuitive use of capability instructions and the thought hadn't even entered my mind. Good one!

mmap.c
```diff
@@ -0,0 +1,102 @@
...
+		0);
+
+	if (result == MAP_FAILED)
+	{
+		printf("ERRNO: %d, ERROR: %s \n\n", errno, strerror(errno));
+	}
+	return result;
+}
+
+uint32_t *generate_purecap(uint32_t *code)
```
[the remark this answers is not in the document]
cristianurlea: See [above comment](https://github.com/capablevms/cheri-examples/pull/1/files#r500235449)

stackscan.c
ltratt: Let's just use the defaults. They might suck, but they suck for everyone equally.

common.h
ltratt: I still didn't understand why it should be a macro or inlined :)

allocate.c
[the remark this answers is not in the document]
ltratt: Works for me!

mmap.c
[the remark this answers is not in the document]
ltratt: I'm not sure that's important in this case? If not, we shouldn't worry about it.

stackscan.c
ltratt: I am struggling to see what advantage these changes would give over the original code. [Which, admittedly, I would write as `address >= base && address <= base + length`, but that's nitpicking on my part.]

common.h
alex4o: Then I would ask do you think this is fine like that?

stackscan.c
cristianurlea: It lowers the cognitive and recall load on (one of) the reviewers, leading to a quicker turn-around time as I won't have to double check operator precedence table. Explicit grouping parentheses also mean i can make use of my code editor to "highlight matching brace". For me personally, the less I have to do in my head, the better.

mmap.c
cristianurlea: I'll resolve current and open separate issue for the column limit.

stackscan.c
ltratt: I think it's a personal coding preference, which is absolutely fine, but I don't think we should push our personal coding preferences onto others in PRs, otherwise we'll never reach a fixed point.

common.h
ltratt: OK, so Aleks and I spoke offline. I now understand the comment, I think. And, assuming I do understand it, I subtly disagree with it :) I agree that `cheri_csp_get` over-approximates the stack, in the sense that you get a pointer that captures more of the stack than you really want. *However* if you were to make this a macro, or inline it, then it's possible that the compiler would decide to move the `asm` statement around, possibly under-approximating the stack -- which would be bad. Now, the compiler probably won't do this, because it tends to be very conservative when it sees `asm` statements, but back when I did a conservative stack scanner in C, I found that having the `asm` statement in a function I "cared" about did lead to subtle bugs.

Long story short, I think the comment should be something more like:

```
// This function returns the current stack pointer.
//
// It might seem that this should be a macro or an inlined function, but then we couldn't be
// entirely sure where the compiler might move the `asm` statement to.
```

That suggests that we should probably mark this function as "never inline".

An alternative approach might be to use compiler barriers to stop reordering. I haven't thought much about those in the context of C, and frankly, it's probably overkill to do so here, but it might be an interesting thing for you to have a look at one day!

mmap.c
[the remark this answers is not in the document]
cristianurlea: I fundamentally disagree, arbitrary enforcement of coding standards dilutes their utility. The large majority of bugs I've found in a live environment got there when they were promoted from "quick hack" to "production code" through code re-use by (likely) someone other than the original author.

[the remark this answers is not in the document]
ltratt: This isn't production code.

stackscan.c
ltratt: I agree that we're still calibrating our comments and that's a perfectly reasonable thing for us to be doing at the moment :)

common.h
cristianurlea: I like the `never inline` approach, it makes the semantics consistent with the minimum level of effort. The  [optnone](https://releases.llvm.org/3.8.0/tools/clang/docs/AttributeReference.html#optnone-clang-optnone) attribute may be of interest in similar contexts.

alex4o: I think I have a new better solution to this problem: 
https://github.com/capablevms/cheri-examples/blob/69dc074e5bd35209b82fce2bb2db478dcc62d368/include/common.h#L30-L36

mmap.c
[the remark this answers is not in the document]
cristianurlea: I won't labour the point further.

mmap.c
alex4o: Fixed in: https://github.com/capablevms/cheri-examples/commit/a37a4e1ce2a506b587a3a5e8c9b4243689576941

include/common.h
```diff
@@ -23,18 +23,15 @@ void inspect_pointer(void *ptr)
...
-// In the current case it gives the stack pointer
-// inside of the cheri_csp_get function.
-// This is fine if only the capabilitits are important.
-// In any other case the csp will need to be decremented
-// to get the actual csp of the calling function.
-void* cheri_csp_get()
+
+// This function returns the current stack pointer.
+//
+// We don't want to inline it because __builtin_frame_address will change it's meaning.
```
ltratt, line 32: I think this line needs expanding a bit. I *think* I know what it's suggesting, but I reckon we need another line or so of explanation to make sure that other idiots (for example, me) can understand what's going on :)

ltratt:
[This comment is just to test MM integration.]

include/common.h
alex4o: I hope this explanation is sufficient.
https://github.com/capablevms/cheri-examples/blob/fdcb452a0716ac7f85acac01bf8ebf24ad54d659/include/common.h#L30-L34

ltratt: Works for me!

ltratt:
And now for the magic words... please squash!

[NB: this doesn't mean you have to squash everything down to a single commit. It just means you've got the opportunity to squash things down if that makes sense.]

cristianurlea:
:tada:

alex4o:
I think It should be ready to merge.

ltratt:
So now I'm going to be a nitpicker and then say something mildly more profound :)

First, can we fix the commit titles as there are some typos? https://github.com/capablevms/cheri-examples/pull/1/commits/92c69fccf4b67cd19507f4c3f2ce9d23b86c914e, https://github.com/capablevms/cheri-examples/pull/1/commits/a812738846418d5732948bda9480e0dbd3a3eeae, and https://github.com/capablevms/cheri-examples/pull/1/commits/a37a4e1ce2a506b587a3a5e8c9b4243689576941

Second, can we add a bit of context to some of the commits? In other words, generally a commit title on its own isn't quite sufficient to make sense when someone looks back at this. Sometimes just a sentence or two in the body is enough; occasionally I end up writing small essays; and sometimes (but not often) nothing is appropriate. I do find that when I do this, I sometimes realise that some commits would make more logical sense squashed together.

Please feel free to do this and then force push an update.

alex4o:
I think this should be sufficient, let me know is something is amiss.

ltratt:
@alex4o I'm not sure you've force pushed?

alex4o:
But I did: https://github.com/alex4o/cheri-examples

ltratt:
> Alright, I had done it, but Github was unable to track it, because I decided to add a fullstop to the first commit and that broke everything. I will rememmber not to touch the root commit from now on.

Ah you mean to the first commit of the repo? Yes, you can only rebase from your branch. `git rebase -i upstream/master` is burnt into my skull as a way of doing this :)

alex4o:
I think these are the last needed changes?

PR #5, opened 2020-10-15
Bump pointer allocation example

jeremysinger:
contribution of simple memory allocator example: a bump pointer allocator.

### DIFF
--- a/example_allocators/bump_allocator/bump_alloc.c
+++ b/example_allocators/bump_allocator/bump_alloc.c
@@ -1,0 +1,78 @@
+/**********************************
+ * bump_alloc.c
+ * Jeremy.Singer@glasgow.ac.uk
+ *
+ * This is a simple bump-pointer allocator.
+ * It mmaps a large buffer of SIZE bytes,
+ * then allocates this space in word-sized
+ * chunks to client code (in main fn).
+ *
+ * Initial simple memory allocator test.
+ * Explore capability narrowing operations
+ * and intrinsics for bound reporting.
+ */
+
+#include <cheriintrin.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+
+#include "bump_alloc.h"
+
+int count = 0;		 /* number of bytes allocated so far*/
+int max = 0;		 /* upper limit for count */
+char *buffer = NULL; /* the allocation buffer */
+
+void init_alloc(int size_in_bytes)
+{
+	/* request memory for our allocation buffer
+	 * NB mmap min bounds for capability is 1 page (4K)
+	 */
+	char *res = mmap(NULL, size_in_bytes, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
+
+	if (res == MAP_FAILED)
+	{
+		perror("error in initial mem allocation");
+		exit(-1);
+	}
+
+	buffer = res;
+	max = size_in_bytes;
+	return;
+}
+
+/*
+ * allocate len bytes with bump pointer allocator
+ * this is our simplistic `malloc` function
+ */
+char *bump_alloc(int len)
+{
+	char *chunk = buffer + count;
+	size_t rounded_len; /* for CHERI alignment */
+	size_t new_count;	/* for buffer overflow check */
+
+	/* ensure we can represent the capability accurately,
+	 * see p30 of CHERI C/C++ Prog Guide (June 2020)
+	 * www.cl.cam.ac.uk/techreports/UCAM-CL-TR-947
+	 */
+	chunk = __builtin_align_up(chunk, ~cheri_representable_alignment_mask(len) + 1);
+	rounded_len = cheri_representable_length(len);
+
+	new_count = (chunk + rounded_len) - buffer;
+	if (new_count > max)
+	{
+		/* out of bounds - don't allocate anything */
+		chunk = 0;
+	}
+	else
+	{
+		/* restrict capability range before returning ptr */
+		chunk = cheri_bounds_set_exact(chunk, rounded_len);
+
+		/* update bytes allocated count */
+		count = new_count;
+	}
+
+	return chunk;
+}
--- a/example_allocators/bump_allocator/Makefile
+++ b/example_allocators/bump_allocator/Makefile
@@ -1,0 +1,28 @@
+CC=$(HOME)/cheri/output/sdk/bin/riscv64-unknown-freebsd13-clang
+
+CFLAGS=-march=rv64imafdcxcheri -mabi=l64pc128d --sysroot=$(HOME)/cheri/output/rootfs-riscv64-hybrid -mno-relax -g -O0 -I../..
+SSHPORT=10003
+export 
+
+cfiles := $(wildcard *.c)
+examples := $(patsubst %.c,bin/%,$(cfiles))
+
+
+good_client: good_client.c bump_alloc.c
+
+bad_client: bad_client.c bump_alloc.c
+
+test: good_client bad_client
+	scp -P $(SSHPORT) good_client bad_client root@127.0.0.1:/root
+	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/good_client'
+	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/bad_client'
+
+bin/%: %.c
+	$(CC) $(CFLAGS) $< -o $@
+
+run-%: bin/%
+	scp -P $(SSHPORT) bin/$(<F) $(<F).c root@127.0.0.1:/root
+	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/$(<F)'
+
+clean: 
+	rm -rfv good_client bad_client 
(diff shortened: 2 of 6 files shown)


ltratt:
I think we've agreed that we should make a couple of tweaks to this PR:

1) Reformat with `clang-format`.
2) Add a "negative" example of the allocator failing so that people can see what happens when CHERI detects errors. This might mean putting the example into its own directory and having a "library" `.c`/`.h` files and a couple of examples which `#include`  the latter?

jeremysinger:
have done the clang-format thing - braces vertically aligned and 8 spaces - urrgh!

example_allocators/bump_allocator/bump_alloc.c
```diff
@@ -0,0 +1,67 @@
...
+char *bump_alloc(int len)
+{
+	char *chunk = 0;
+	if (count + len <= max)
+	{
+		/* bump the pointer */
+		chunk = buffer + count;
+
+		/* restrict capability range before returning ptr */
+		chunk = cheri_bounds_set_exact(chunk, len);
```
jacobbramley, line 60: Did you give any thought to how you might give the user _at least_ as much memory as they requested? Aborting with awkward lengths is fine for an example, but not ideal for a real allocator. Simply using `cheri_bounds_set` (not `_exact`) isn't the answer because (if I understand correctly) that might adjust the base downwards, into already-allocated memory.

ltratt:
@jeremysinger Is this ready for rereview?

example_allocators/bump_allocator/bump_alloc.c
jacobbramley: Oh, the CHERI C/C++ Programming Guide has a section (7.5) on this, with an example implementation.

jeremysinger: thanks @jacobbramley I have added this fixup code now for alignment ... please could you review?

example_allocators/bump_allocator/bump_alloc.c
```diff
@@ -51,16 +51,23 @@ void init_alloc(int size_in_bytes)
...
 		chunk = buffer + count;
 
+		/* check we can represent the capability accurately,
+		 * see p30 of CHERI C/C++ Prog Guide */
+		chunk = __builtin_align_up(chunk, ~cheri_representable_alignment_mask(len) + 1);
+		rounded_len = cheri_representable_length(len);
+
 		/* restrict capability range before returning ptr */
-		chunk = cheri_bounds_set_exact(chunk, len);
+		chunk = cheri_bounds_set_exact(chunk, rounded_len);
```
jacobbramley, line 67: To be clear, this will crash if there isn't space in `chunk`'s bounds. I think that's Ok here, and in many bump-pointer applications, but a production allocator will probably want to handle it explicitly (e.g. returning NULL, like malloc).

example_allocators/bump_allocator/bump_alloc.c
```diff
@@ -51,16 +51,23 @@ void init_alloc(int size_in_bytes)
...
 
 		/* update bytes allocated count */
-		count += len;
+		count += rounded_len;
```
jacobbramley, line 70: Just an observation on usage (not a problem with the example): the caller should query the capability length, rather than assume that the buffer has exactly `len` bytes, for example if they want to zero the buffer or otherwise poison it. The [VIXL W^X example](https://git.linaro.org/arm/vixl.git/tree/examples/aarch64/morello-wx.cc?h=morello#n132) pads with known-undefined instructions because I think that's what a JIT would do as a security precaution.

jacobbramley (approved):
Looks good to me!

My comments are to start discussions only :-)

example_allocators/bump_allocator/bump_alloc.c
```diff
@@ -51,16 +51,23 @@ void init_alloc(int size_in_bytes)
...
 	char *chunk = 0;
+	size_t rounded_len; /* for CHERI alignment */
+
 	if (count + len <= max)
 	{
 		/* bump the pointer */
 		chunk = buffer + count;
 
+		/* check we can represent the capability accurately,
+		 * see p30 of CHERI C/C++ Prog Guide */
```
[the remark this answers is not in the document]
ltratt: We could maybe give the version number of the guide used?

example_allocators/bump_allocator/bump_alloc.c
ltratt: >To be clear, this will crash if there isn't space in chunk's bounds

Are you saying that at least one problem is that `rounded_len` will over-approximate the memory so we'll end up providing bigger bounds than we should? Or something else as well?

example_allocators/bump_allocator/bump_alloc.c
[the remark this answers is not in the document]
jeremysinger: good idea  -I'll point to the UCAM Tech Report number and its date - this shouldn't change (I think )

example_allocators/bump_allocator/bump_alloc.c
jeremysinger: oops! - I think you are right - I only check we have enough space for len bytes, not enough space for rounded_len bytes ... that's very silly of me. Will fix up now :-)

example_allocators/bump_allocator/bump_alloc.c
jacobbramley, line 70: Oh, i nearly missed it: I think we also need to bump by the amount that we aligned up.

`count += (chunk' - chunk) + rounded_len`

example_allocators/bump_allocator/bump_alloc.c
ltratt: Well, here's an interesting question: do we *want* to allocate the `rounded_len` number of bytes because that will often waste some memory, potentially mega/gigabytes. So one way one to look at the capability is that it guarantees to cover all of our heap, but possibly also a lot of things beyond the heap too. But that might still be better than not having a capability at all?

jacobbramley: I think I meant `buffer`, sorry.

There are (at least) three interesting cases:
1. The caller asks for more memory than is actually available, so `len > length(buffer)`.
2. `len` is large and unrepresentable so we have to round it up.
3. The base (at `buffer + count`) is insufficiently aligned for the `rounded_len`.

Case 1 is obviously an out-of-memory condition.

I don't think cases 2 and 3 can cause an error by themselves (in the absence of case 1). We pad at one end of the allocation, but I don't think that padding can put `chunk`'s limit outside `buffer` because (in the absence of 1) `len <= length(buffer)` and so it has at least as fine a precision. This is intuition at best; a proof would be useful, I think.

The combination of 2 and 3 can pad at both ends, and that can make an allocation fail even if there are enough actual bytes remaining. It'll be hard to hit (and therefore test) in practice.

jacobbramley: @ltratt I think that's a per-allocator policy decision. If permissions will differ (e.g. the W^X example) then you need to, to prevent overlaps that could be security issues. In other situations (especially if allocations will be further sub-divided anyway) then it may be acceptable.

The wastage may be mega/gigabytes but still a very small compared to the allocation, and much larger than hardware pages (so it's just VA space, not physical memory).

ltratt: Point taken!

example_allocators/bump_allocator/bump_alloc.c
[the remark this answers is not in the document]
jeremysinger: Done - UCAM-TR-947 dated June 2020 - this isn't going to change, ever...

example_allocators/bump_allocator/bump_alloc.c
jeremysinger: hi @jacobbramley thanks for the detailed feedback. I am now considering all three interesting cases you mention above. Previously, I was unaware of case (3) so this was a useful exercise.

example_allocators/bump_allocator/bump_alloc.c
jeremysinger: done - thanks for feedback...

example_allocators/bump_allocator/bump_alloc.c
```diff
@@ -0,0 +1,80 @@
...
+	char *res = mmap(NULL, size_in_bytes, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
+
+	if (res == MAP_FAILED)
+	{
+		perror("error in initial mem allocation");
+		exit(-1);
+	}
+
+	/* NB mmap min bounds for capability is 1 page (4K) */
+	// if (DEBUG_PRINTF)
```
ltratt, line 39: We should probably remove commented code?

ltratt:
This looks good to me, so if @jacobbramley is OK with it, I think we'll start moving towards merging!

example_allocators/bump_allocator/bump_alloc.c
jeremysinger: fixed in 6b7294a

jacobbramley:
Looks good to me too!

ltratt:
@jeremysinger Please squash as you see fit (you don't have to squash down to a single commit, though you can if you want!).

jeremysinger:
> @jeremysinger Please squash as you see fit 

I think I have squashed everything now ... is that ok?

jeremysinger:
hi @ltratt I think this is the tidied up, squashed atomic commit of the bump pointer example. Has this saga almost reached an end? 😄

PR #8, opened 2020-11-03
Enable c++

alex4o:
Allow for c++ code to be used in the examples, and switch to purecap mode for the sysroot.
Also, switch to the intrinsics in all the examples to the ones from the sysroot as their header file work with c++.
The reason for the switch is that on the CHERI slack it was mentioned that the purecap mode should be the default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
 CC=$(HOME)/cheri/output/sdk/bin/riscv64-unknown-freebsd13-clang
+CXX=$(HOME)/cheri/output/sdk/bin/riscv64-unknown-freebsd13-clang++
 CFLAGS=-march=rv64imafdcxcheri -mabi=l64pc128d --sysroot=$(HOME)/cheri/output/rootfs-riscv64-purecap -mno-relax -g -O0
 SSHPORT=10021
 export 
 
 cfiles := $(wildcard *.c)
-examples := $(patsubst %.c,bin/%,$(cfiles))
+cppfiles := $(wildcard *.cpp)
+examples := $(patsubst %.c,bin/%,$(cfiles)) $(patsubst %.cpp,bin/%,$(cfiles))
 
 .PHONY: all run clean
 
@@ -12,6 +14,10 @@ all: $(examples)
 
 bin/%: %.c
 	$(CC) $(CFLAGS) $< -o $@
+	scp -P $(SSHPORT) $< root@127.0.0.1:/root
+
+bin/%: %.cpp
+	$(CXX) $(CFLAGS) $< -o $@
 	scp -P $(SSHPORT) $< root@127.0.0.1:/root
 
 run-%: bin/%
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 CC=$(HOME)/cheri/output/sdk/bin/riscv64-unknown-freebsd13-clang
-CFLAGS=-march=rv64imafdcxcheri -mabi=l64pc128d --sysroot=$(HOME)/cheri/output/rootfs-riscv64-hybrid -mno-relax -g -O0
-SSHPORT=10017
+CFLAGS=-march=rv64imafdcxcheri -mabi=l64pc128d --sysroot=$(HOME)/cheri/output/rootfs-riscv64-purecap -mno-relax -g -O0
```
ltratt, line 3: Can all examples be built in `purecap` mode?

alex4o: Yes they do.  Buildbot also supports purecap mode so this shouldn't be a problem.

ltratt: OK, so dumb question: why did we specify `hybrid` before?

alex4o: Because we/I didn't know any better, and somewhere it was written that purecap is broken so the hybrid one should be used.

ltratt: Point taken!

setjmp.c
```diff
@@ -18,12 +19,12 @@ int main()
...
-		if (cheri_is_valid(((void **)buffer)[idx]))
+		if (cheri_gettag(((void **)buffer)[idx]))
 		{
 			void *csp = cheri_csp_get();
 			uint64_t base = cheri_base_get(csp);
 			uint64_t length = cheri_length_get(csp);
-			uint64_t address = cheri_address_get(((void **)buffer)[idx]);
+			uint64_t address = cheri_getaddress(((void **)buffer)[idx]);
 
 			if (address >= base && address <= (base + length))
```
cristianurlea, line 29: Only just noticed this even though it's not a change:  The upper bound is [not inclusive](https://github.com/CTSRD-CHERI/newlib/blob/d92070d2367ad72dcbf161825abdc64a7a1ec4f0/newlib/libc/include/cheric.h#L90) so you might want to tweak it to `address < (base + length)`

stackscan.c
```diff
@@ -15,11 +15,11 @@ void inspect_stack(void *stack)
...
 {
-	if (cheri_is_valid(ptr))
+	if (cheri_gettag(ptr))
 	{
 		uint64_t base = cheri_base_get(stack_top);
 		uint64_t length = cheri_length_get(stack_top);
-		uint64_t address = cheri_address_get(ptr);
+		uint64_t address = cheri_getaddress(ptr);
 
 		if (address >= base && address <= (base + length))
```
cristianurlea, line 24: Same issue [as above](https://github.com/capablevms/cheri-examples/pulls#r516624074)

alex4o: Fixed.

setjmp.c
alex4o: Fixed.

alex4o:
I think this can be relaiably merged now.

ltratt:
Can we extend the commit message for https://github.com/capablevms/cheri-examples/pull/8/commits/7808bb539aa5a4c3e1563e250eeb2915ba3c3d5d to explain *why* we're moving to purecap mode? Please feel free to force push an updated commit message.

alex4o:
Done

alex4o:
bors try

ltratt:
Please squash.

ltratt:
Is this ready?

alex4o:
Done

PR #9, opened 2020-11-03
Add a sentry example.

alex4o:
The main idea behind sentries is that they protect the function pointers from being modified, read to/from. This allows emulation of most sealed capability modes. This example shows some of the use cases of sentry capabilities that I could think of. This is kind of strange to understand because CHERI switched to using sentry capabilities everywhere.

### DIFF
--- a/sentry.c
+++ b/sentry.c
@@ -1,0 +1,169 @@
+/*
+ * An example showing how a sentry can be used for more advanced
+ * security model. In CHERIv8 every function is actually a sentry.
+ * However sentries can allow for additional security when used with
+ * a level of indirection. In this example an assembly trampoline is
+ * used to load additional data, that isn't available to the user.
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <err.h>
+
+#include "include/common.h"
+#include "include/instructions.h"
+#include "include/regs.h"
+
+#include <cheri/cheric.h>
+
+struct data
+{
+	int a;
+	int b;
+	char c[256];
+};
+
+void *setup_sentry(void *, void (*f)(uint32_t *, void *));
+void oop_sentry(struct data *, int);
+void simple_sentry(int);
+
+// This is the function that will be called through a sentry and
+// the first argument is passed by the trampoline assembly.
+void oop_sentry(struct data *this, int arg)
+{
+	printf("OOP: %p %d \n", this, arg);
+}
+
+// This function will also be called through a sentry but no
+// arguments are added.
+void simple_sentry(int arg)
+{
+	printf("Simple: %d \n", arg);
+}
+
+// A function to generate the code that will put the
+// this structure as first argument and move the rest of the
+// arguments to the next registers.
+void gen_oop(uint32_t *code, void *function)
+{
+	intptr_t *code_data = (intptr_t *)code;
+
+	// The additional data needs to be stored somewhere in memory
+	// where it can not be accessed by the user of the function.
+	// This is why it is stored in the same place where the code will be.
+	// As the code is not modifiable and not readable this data is also
+	// protected.
+	code_data[255] = (intptr_t)function;
+	code_data[254] = (intptr_t)malloc(sizeof(struct data));
+
+	uint32_t idx = 0;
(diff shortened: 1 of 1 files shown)


sentry.c
```diff
@@ -0,0 +1,124 @@
...
+	code[idx++] = addi(a1, a0, 0);
+	code[idx++] = addi(a2, a1, 0);
+	code[idx++] = addi(a3, a2, 0);
+	code[idx++] = addi(a4, a3, 0);
+	code[idx++] = addi(a5, a4, 0);
+	code[idx++] = addi(a6, a5, 0);
+	code[idx++] = addi(a7, a6, 0);
+	code[idx++] = clc_128(ca0, cs2, ((-32) + (2 << 20)));
+	code[idx++] = cjalr(cnull, cs3);
+
```
ltratt, line 47: Extraneous blank lines.

ltratt:
I think it might be worth documenting somewhere (maybe at the top of the file?) what the intended purpose of this example is. I get the rough idea, but I'll have forgotten by tomorrow :)

cristianurlea:
A small description of what each function does would be helpful ( as a doc/comment on top of each ).  This is especially important when most of the functionality is assembled inline and thus obscured to mortals.

sentry.c
cristianurlea: It pains me to ask for this, as I prefer `type* name` over `type *name`, but this needs to be passed by `clang-format` as well.

sentry.c
```diff
@@ -0,0 +1,124 @@
...
+			fpointer = setup_sentry(simple_sentry, gen_override);
+			break;
+		case 3:
+			fpointer = setup_normal(simple_sentry, gen_simple);
+			break;
+		case 4: // This case should fail
+			fpointer = setup_sentry(simple_sentry, gen_simple);
+			fpointer = cheri_incoffset(fpointer, 16);
+			break;
+	}
```
cristianurlea, line 120: Adding the following case would document that you have considered the fall-through case and decided against giving it special behaviour. 
```
 default :  // do nothing;
              break; 
```

sentry.c
```diff
@@ -0,0 +1,124 @@
...
+
+	fpointer(99);
```
cristianurlea, line 122: Call to potentially uninitialised `fpointer` (if `mode` not \in [0..4] )

sentry.c
ltratt: I agree that a `default` case here might be a good idea, but I think the right thing to do would be to `err` and exit immediately if it happens?

cristianurlea: > I think the right thing to do would be to `err` and exit

Given the call to `fpointer` below that is almost certainly the case.

An alternative hazard removal strategy is to restrict the scope of `fpointer` to each case. This has the added benefit of making data-flow a analysis easie. There would be nothing left to make LLVM ponder the actual state of `fpointer` past the switch statement. This is especially important given the calls to cheri intrinsics (that can have effects on `fpointer`) which make the eventual state of `fpointer` opaque. 

There are also potential failure from `mmap` to account for that may influence the author's course of action here.

alex4o: I have decided to make the default case error and exit:
https://github.com/capablevms/cheri-examples/blob/ecf35f220f6e91e3a187ced602ab67e9da616ea0/sentry.c#L153-L155

sentry.c
alex4o: I should add a commit hook to format before pushing.

alex4o:
Force pushed.

bors try

sentry.c
ltratt: Isn't that the same as `err(1, "Invalid...")` (I think exit codes are conventionally `>0`)?

alex4o: I didn't know about the `err` function. I am using it now.

cristianurlea: You should use `EXIT_FAILURE` from `<stdlib.h>`. Generally a return code  `> 0` indicates failure but that is not portable. ( /*disapprovingly stares* at VMS )

ltratt:
Please squash.

alex4o:
bors try

ltratt:
Is this ready?

alex4o:
Yes

PR #12, opened 2020-12-01
Add example showing the usage of .so files.

alex4o:
This example is needed to present the implications of using shared libraries. How they can be used as a proxy for deciding what is
public and what is private by means of what is being exported and how does it integrate with the changes made on the linker (rtld) as needed by CHERI. 
This is what lead to deciding to split the lua VM into multiple shared objects. But this example also contains the idea for creating compartments purely from the program counter capability (PCC) and using it to protect data that should be private for a function.

### DIFF
--- a/shared_objects/Makefile
+++ b/shared_objects/Makefile
@@ -1,0 +1,16 @@
+CC=$(HOME)/cheri/output/sdk/bin/riscv64-unknown-freebsd13-clang
+CFLAGS=-march=rv64imafdcxcheri -mabi=l64pc128d --sysroot=$(HOME)/cheri/output/rootfs-riscv64-hybrid -mno-relax -g
+SSHPORT=10021
+
+main: main.c
+	$(CC) $(CFLAGS) -L. -l1 -l2 -g $< -o main
+
+lib1.so: lib1.c
+	$(CC) $(CFLAGS) -g -fPIC -shared $< -o lib1.so
+
+lib2.so: lib2.c
+	$(CC) $(CFLAGS) -g -fPIC -shared $< -o lib2.so
+
+run:
+	scp -P $(SSHPORT) * root@127.0.0.1:/root/demo/
+	ssh -p $(SSHPORT) root@127.0.0.1 -t 'setenv LD_LIBRARY_PATH /root/demo; /root/demo/main'
--- a/shared_objects/lib1.c
+++ b/shared_objects/lib1.c
@@ -1,0 +1,25 @@
+/**
+ * A basic example showing how CHERI sealed entries can be used to protect data. In this case the
+ * cont variable will be inaccessible from the things that link with this shared library.
+ */
+
+#include "include/lib1.h"
+
+static int32_t count = -5;
+
+void increment()
+{
+	count += 1;
+}
+
+int get_count()
+{
+	if (count > 0)
+	{
+		return count;
+	}
+	else
+	{
+		return 0;
+	}
+}
--- a/shared_objects/lib2.c
+++ b/shared_objects/lib2.c
@@ -1,0 +1,70 @@
+/**
+ * A more complicated example of sealed entries displaying how they can be used to protect private
+ * data that will and can be given to the user of the library as a pointer. In this case Car_priv is
+ * the private fields and methods of the object that is given to the user, and it is protected by
+ * exploiting the properties of the sealed entries (The fact that the capability can only be jumped
+ * to not modified in any way).
+ */
+
+#include <cheri/cheri.h>
+#include <cheri/cheric.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <stddef.h>
+
+#include "../include/instructions.h"
+#include "../include/regs.h"
+#include "include/lib2.h"
+
+static struct Car *arena;
+static uint32_t size;
+
+void init() {}
+
+void crash()
+{
+	printf("CRASH!\n");
+}
+
+void honk(struct Car_priv *priv, struct Car *car)
+{
+	printf("%p %p\n", priv, car);
+	printf("HONK!\n");
+	if (car->speed > priv->maxSpeed)
+	{
+		priv->crash();
+	}
+}
+
+struct Car *new_car()
+{
+	uint32_t *mem =
+		mmap(NULL, 4096, PROT_READ | PROT_WRITE | PROT_EXEC, MAP_ANON | MAP_PRIVATE, -1, 0);
+
+	void **ptrs = mem;
+	
+	const uint32_t vtable_start_index = 2;
+	ptrs[vtable_start_index + 0] = honk;
+
+	const ptrdiff_t vtable_offset = ((char*)(ptrs + vtable_start_index) - (char*)ptrs);
+	const size_t vtable_size = 1 * 16;
+
+	uint32_t idx = 0;
+	mem[idx++] = auipcc(cs2, 0);
+	mem[idx++] = clc_128(cs3, cs2, vtable_offset);
+	mem[idx++] = cincoffsetimm(ca0, cs2, vtable_offset + vtable_size);
+	mem[idx++] = cincoffsetimm(ca1, cs2, vtable_offset + vtable_size + sizeof(struct Car_priv));
+	mem[idx++] = cjalr(cnull, cs3);
+
+	uint32_t functions_size = vtable_offset + vtable_size;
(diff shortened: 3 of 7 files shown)


ltratt:
Can we add a comment or two explaining what these are showing? I think they alson eed formatting.

shared_objects/main.c
```diff
@@ -0,0 +1,27 @@
+#include<stdio.h>
+#include<stdint.h>
+#include<stdlib.h>
+
+#include "include/lib1.h"
+#include "include/lib2.h"
+
+int main() <%
```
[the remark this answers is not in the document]
alex4o: I found that <% can be used instead of { and tried it, but I forgot to remove it.

shared_objects/Makefile
```diff
@@ -0,0 +1,17 @@
+CC=$(HOME)/cheri/output/sdk/bin/riscv64-unknown-freebsd13-clang
+CFLAGS=-march=rv64imafdcxcheri -mabi=l64pc128d --sysroot=$(HOME)/cheri/output/rootfs-riscv64-hybrid -mno-relax -g
+SSHPORT=10021
+export
```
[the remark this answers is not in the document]
alex4o: I can see where?

alex4o:
Done

shared_objects/Makefile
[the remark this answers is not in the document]
ltratt: I assumed `export` on its own line wasn't supposed to be there? But maybe it does something I don't know about?!

shared_objects/lib1.c
```diff
@@ -1,3 +1,11 @@
+/**
+ * A basic example showing how CHERI
```
ltratt: Can we format these to a slightly wider width -- 76, 80, or 100 chars are all reasonable choices IMHO, but does waste a lot of horizontal screen space in these days of widescreen monitors :/

shared_objects/Makefile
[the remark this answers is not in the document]
alex4o: Oh yes it is supposed to make all the variables before it be stable from the env, like `CC=gcc make`.

shared_objects/lib1.c
alex4o: done

shared_objects/Makefile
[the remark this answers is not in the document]
ltratt: Is it needed here? I think it has the same effect with and without?

ltratt:
@alex4o Let's choose a consistent comment width, rather than the random widths of https://github.com/capablevms/cheri-examples/pull/12/commits/4ff9e9bfe66298f712ec4e0697896efa36e0780b :)

alex4o:
Yes sry, I should have relied on clang-format, to begin with.

ltratt:
Please squash.

shared_objects/Makefile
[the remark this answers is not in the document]
alex4o: The reason it is here is because I copied that from the root Makefile, which originally I wanted to be capable of running make in the subdirectories and pass the CC and CFLAGS from there. You are right it might not be needed.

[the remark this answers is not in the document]
ltratt: If it isn't needed, let's kill it, otherwise we'll keep copy and pasting it without ever understanding it :) If it's also unnecessary at the top level, you could do a separate PR which just kills it there too?

shared_objects/lib2.c
```diff
@@ -0,0 +1,61 @@
...
+	uint32_t idx = 0;
+	mem[idx++] = auipcc(cs2, 0);
+	mem[idx++] = clc_128(cs3, cs2, 32);
+	mem[idx++] = cincoffsetimm(ca0, cs2, 32 + 16);
+	mem[idx++] = cincoffsetimm(ca1, cs2, 32 + 16 + sizeof(struct Car_priv));
+	mem[idx++] = cjalr(cnull, cs3);
+	void **ptrs = mem;
+	ptrs[2] = honk;
+
+	uint32_t functions_size = 32 + 16;
```
cristianurlea, line 51: Is there any providence we can make explicit for the magic literals: `32` and `16` here and above ?

ltratt:
@alex4o Ping.

shared_objects/lib2.c
alex4o: Not really, it is the size of the instructions + the size of the functions pointers. I will add a comment about that.

alex4o:
Removed the export and added a comment.

shared_objects/lib2.c
```diff
@@ -42,7 +42,9 @@ struct Car *new_car()
 	uint32_t idx = 0;
 	mem[idx++] = auipcc(cs2, 0);
 	mem[idx++] = clc_128(cs3, cs2, 32);
-	mem[idx++] = cincoffsetimm(ca0, cs2, 32 + 16);
+	mem[idx++] = cincoffsetimm(ca0, cs2, 32 + 16); 
+	// 32 is the size of the stubs and 16 is the size of the function pointers in bytes
```
ltratt, line 46: I think the comment should go above the first line referencing `32` and `16`. An obvious question, though: can we make these constants based on some sort of "size of" calculation rather than hard coding them?

alex4o: I would have used size of if I could, but there is no way to express the size of these things or I just don't know about it.

ltratt: OK, a comment explaining that it's a bit difficult to calculate and also what platforms the calculation is likely to be correct on would be cool.

alex4o: This should be better now then?

ltratt: No "also what platforms the calculation is likely to be correct on"?

ltratt:
@alex4o That last commit isn't showing up from your normal email address?

alex4o:
Yes, I saw that and fixed it. There should be no problem when I squash.

ltratt:
OK.

shared_objects/lib2.c
alex4o: This should also be addressed now.

ltratt:
Please squash.

alex4o:
Done, buildbot should work today.

ltratt:
Is https://github.com/capablevms/cheri-examples/pull/12/commits/d8775f9e6340dcce74934a666ca85cd9ac298083 new and needs reviewing?

cristianurlea:
`d8775f9` addresses the point raised [here](https://github.com/capablevms/cheri-examples/pull/12#discussion_r535339223) and [here](https://github.com/capablevms/cheri-examples/pull/12#discussion_r536013107) by making the code self-documenting. Aside from the minor inline commented of  /s/`off_t` /`uintptr_t` I think this removes any and all magic constant mysticism from the example.

alex4o:
> `d8775f9` addresses the point raised [here](https://github.com/capablevms/cheri-examples/pull/12#discussion_r535339223) and [here](https://github.com/capablevms/cheri-examples/pull/12#discussion_r536013107) by making the code self-documenting. Aside from the minor inline commented of /s/`off_t` /`uintptr_t` I think this removes any and all magic constant mysticism from the example.

So I checked the [Cheri C/C++ programming guide](https://github.com/CTSRD-CHERI/cheri-c-programming) (You need to build it yourself for the newest version). And it says: 
```
These integer types should be used to hold values that may be validpointers if cast back to a pointer type. 
When anintptr_tis assigned an integer value– e.g., due to constant initialization to an integer in the source
 – and the result is cast to apointer type, the pointer will be invalid and hence non-dereferenceable.
```
The way I understand this, is that it should hold the tag too, which isn't something that we care about?

cristianurlea:
> The way I understand this, is that it should hold the tag too, which isn't something that we care about?

Indeed we don't. I think i found a more suitable type, namely `ptrdiff_t` 

``
[ Note: Should we recommend that size_t be used to hold lengths of allocations and
ptrdiff_t be used to talk about spans of address space (e.g., the offsets between two
subobjects of an allocation)? I feel like the recommendations here are not as concrete as
I’d like. - nwf]
``

alex4o:
Oh yes you are correct I should have seen that earlyer.

ltratt:
Agreed, `ptrdiff_t` is more appropriate.

Please sqash.

alex4o:
Done, please try bors.

ltratt:
bors try

ltratt:
bors try-

PR #13, opened 2021-01-28
Added a third shared library example and organize the code.

alex4o:
The purpose of the third example is to show, which functions are available to call from each function, and how that changes. As CHERI is using the PCC as a reference of the place where the PLT resides. Thus scanning the PCC reveals all of the accessible functions (and also static variables). 

Side note: I realized that CHERI gives us something akin to reflection for C, which is interesting and might be worth looking into.


ltratt:
> The purpose of the third example is to show, which functions are available to call from each function, and how that changes. As CHERI is using the PCC as a reference of the place where the PLT resides. Thus scanning the PCC reveals all of the accessible functions (and also static variables).

I'm not sure I understand this. Is the example program expected to succeed or fail or ... ? A comment at the top of the file explaining what it's doing is almost certainly going to be a good idea.

alex4o:
It is supposed to succeed, and show you how things look like in practice.

ltratt:
Shouldn't we also add a second *negative* example that shows what happens if you access a function outside of your compartment?

alex4o:
It turn's out I had forgotten to add the new example in. So the new commit adds the example with the relevant description.

shared_objects/unexported_function.c
```diff
@@ -0,0 +1,30 @@
...
+#include "include/unexported_function.h"
+#include "include/find_sentries.h"
+
+__attribute__ ((visibility ("hidden")))
+void do_work() {
+	printf("Doing work\n");
+}
+
+int test() {
+
```
ltratt, line 19: Many blank lines!

ltratt:
At the moment these are interesting examples for humans, in the sense that you can run them, but I wonder if we can make them *tests*: i.e. check that we can't reference / call / whatever another function? I think that would make their intent clearer.

shared_objects/unexported_function.c
alex4o: I removed them, but I very much like my blank lines.

ltratt: :p

alex4o:
Done it is an automatic test using asserts now. And the printing of the table makes it clearer what you need to spot.

ltratt:
Please squash.

alex4o:
I didn't squash them, but I reworded them.

PR #14, opened 2021-02-10
cmake build file with riscv toolchain file

djichthys:
* Use CMake to build all the cheri examples
* Select architecture to compile for from command line 
* Set tool chain and SDK paths based on architecture

*Note* : currently only the risc-v implementation is done

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,29 @@
 # cheri-examples
 cheri-riscv sample c programs
+
+------
+## Build instructions with CMake
+Build all the example code with CMake using default settings :
+1. Create a separate build directory (\<build-dir\>) outside the source directory (\<source-dir\>).
+2. Generate the build files using either **Unix Makefiles** (*default*) or **Ninja** (specfied
+   with **-G**). Currently only the *RISC-V* cheri-purecap toolchain file has been created.
+   1. If executing from inside the build directory :
+
+   `cmake -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake <relative-path-to-source-dir>`
+
+   2. If executing from outside the build directory :
+
+   `cmake -B <build-dir> -S <source-dir> -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake`
+
+3. Build all the examples
+   1. If executing from inside the build directory,  `make all` *OR* `ninja all`
+   2. If executing from outside the build directory, `cmake --build <build-dir>`
+
+#### CMake options:
+* **-DCMAKE_TOOLCHAIN_FILE**: Use `-DCMAKE_TOOLCHAIN_FILE=<toolchain-file>` to select
+  architecture to compile binary for. Only *riscv64-purecap.cmake* is currently supported.
+* **-DSDK**: Use `-DSDK=<source-to-sdk>` to point to the *CHERI* SDK directory.
+  (Default path: ${HOME}/cheri/output/sdk)
+* **-G**: Choose preferred build system
+  - Use `-G "Unix Makefiles"` to build using *makefiles* (*default*).
+  - Use `-G Ninja` to build using *ninja*.
--- a/riscv64-purecap.cmake
+++ b/riscv64-purecap.cmake
@@ -1,0 +1,23 @@
+set(CMAKE_SYSTEM_NAME Linux)
+
+# Default SDK path
+set(SDK "$ENV{HOME}/cheri/output/sdk" CACHE PATH "path to cheri SDK")
+
+# Set toolchain compilers
+set(CMAKE_C_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-cc)
+set(CMAKE_CXX_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-c++)
+
+# Don't run the linker on compiler check
+set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+
+# Define sysroot path for CHERI-build`
+set(CMAKE_SYSROOT ${SDK}/sysroot-riscv64-purecap)
+
+# Use only cross compiler tools for compilation and linking
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+
+# Set correct machine and abi flags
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,2 +1,11 @@
 # cheri-examples
 cheri-riscv sample c programs
+
+## Build instructions with CMake
+1. Create a separate build directory (\<build-dir\>) outside the source directory (\<source-dir\>).
+2. Generate the build files using either **Unix Makefiles** or **Ninja**. Currently only the *RISC-V* cheri-purecap toolchain file has been created. From outside the source directory execute:
```
ltratt, line 6: We need to wrap long lines (in general: either to 80 or 100 chars as you prefer; but if the file/repo has already set a standard, let's stick with it), to help `diff` in the future (and our future selves too!).

README.md
```diff
@@ -1,2 +1,11 @@
 # cheri-examples
 cheri-riscv sample c programs
+
+## Build instructions with CMake
+1. Create a separate build directory (\<build-dir\>) outside the source directory (\<source-dir\>).
+2. Generate the build files using either **Unix Makefiles** or **Ninja**. Currently only the *RISC-V* cheri-purecap toolchain file has been created. From outside the source directory execute: 
+
+   `cmake -B <build-dir> -S <source-dir> -DCMAKE_TOOLCHAIN_FILE=<source-dir>/riscv64-purecap.cmake -DSDK=<cheri-sdk-dir> -G Ninja`
```
ltratt, line 8: Can we simplify this in some way? It is very complex!

riscv64-purecap.cmake
```diff
@@ -0,0 +1,23 @@
...
+#Set toolchain compilers
+set(CMAKE_C_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-cc)
+set(CMAKE_CXX_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-c++)
+
+
+# Don't run the linker on compiler check
+set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+
+#Define sysroot path for CHERI-build`
+#set(CMAKE_FIND_ROOT_PATH ${SDK}/sysroot-riscv64-purecap)
```
ltratt, line 12: In general, we don't want to include commented code (unless there's a good reason) -- I guess this is a left-over piece of debugging?

riscv64-purecap.cmake
```diff
@@ -0,0 +1,23 @@
+set(CMAKE_SYSTEM_NAME Linux)
+
+#Set toolchain compilers
```
ltratt, line 3: I tried to suppress my OCD, but I have failed :) Can we consistently use a space character after the `#`? :) Also, we're sometimes using one, and sometimes two, newlines between sections, and I wasn't sure if that implied bigger "semantic" differences or not -- I *think* not, but I might be wrong!

README.md
djichthys: Added defaults to *cmake* config file to reduce need for command-line options in most cases.

README.md
```diff
@@ -1,11 +1,29 @@
...
+Build al the example code with CMake using  default settings :
 1. Create a separate build directory (\<build-dir\>) outside the source directory (\<source-dir\>).
-2. Generate the build files using either **Unix Makefiles** or **Ninja**. Currently only the *RISC-V* cheri-purecap toolchain file has been created. From outside the source directory execute: 
+2. Generate the build files using either **Unix Makefiles** (*default*) or **Ninja** (specfied
+   with **-G**). Currently only the *RISC-V* cheri-purecap toolchain file has been created.
+   1. If executing from inside the build directory :
 
-   `cmake -B <build-dir> -S <source-dir> -DCMAKE_TOOLCHAIN_FILE=<source-dir>/riscv64-purecap.cmake -DSDK=<cheri-sdk-dir> -G Ninja`
-3. Build all the examples 
+   `cmake -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake <relative-path-to-source-dir>`
```
ltratt, line 12: Can we make `CMAKE_TOOLCHAIN_FILE` a default too?

djichthys: I don't think we can use defaults for setting the tool chain file. If we set a default,   it will have to be done before the [`project()`](https://github.com/djichthys/cheri-examples/blob/6478241cd5c949fa75e58679c7b167c826029634/CMakeLists.txt#L3)  directive to take effect.  However, cmake then does not seem to be able to override it when we use a different file (say morello).

README.md
```diff
@@ -1,11 +1,29 @@
 # cheri-examples
 cheri-riscv sample c programs
 
+------
 ## Build instructions with CMake
+Build al the example code with CMake using  default settings :
```
ltratt, line 6: s/al/all/ ? Also notice the double space before "default".

README.md
```diff
@@ -1,11 +1,29 @@
...
+2. Generate the build files using either **Unix Makefiles** (*default*) or **Ninja** (specfied
+   with **-G**). Currently only the *RISC-V* cheri-purecap toolchain file has been created.
+   1. If executing from inside the build directory :
 
-   `cmake -B <build-dir> -S <source-dir> -DCMAKE_TOOLCHAIN_FILE=<source-dir>/riscv64-purecap.cmake -DSDK=<cheri-sdk-dir> -G Ninja`
-3. Build all the examples 
+   `cmake -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake <relative-path-to-source-dir>`
 
-   `cmake --build <build-dir>`
+   2. If executing from outside the build directory :
```
ltratt, line 14: Will anyone ever want to build from outside the build directory? This feels like it might come under the "if you want to do this, you know what to do anyway" heading?

README.md
djichthys: Guilty! Fixed. will push them soon.

README.md
djichthys: Keeps it consistent with make/ninja rule generation instructions. Let me know if not necessary and I can remove it

ltratt: I have no preferences here, as I don't use systems like ninja. So if you think that's best, let's go with it.

ltratt:
@djichthys ping

djichthys:
Hi @ltratt . I think this PR is ready unless there are other comments.

ltratt:
Please squash (this doesn't mean you have to squash down to one commit; but you have the opportunity to squash away any commits you think don't add anything useful to history).

djichthys:
Force pushed after squashing to a single commit. Please do let me know if I have done something stupid :)

ltratt:
Can we add some more details to the commit message, specifically about why we've done this (not just that we've done it :))? Feel free to force push an update.

djichthys:
Bigger description in commit message

PR #15, opened 2021-02-22
Integrate the new compilation process, and add the cmake build to the .buildbot.sh build script.

alex4o:
Switched from using `riscv64-unknown-freebsd13-clang` to
`clang --config cheribsd-riscv64-purcap.cfg`. Now the only
linker that can be used is llvm's lld by doing `-fuse-ld=lld`.
Also included the new cmake build system in the build script.

### DIFF
--- a/riscv64-purecap.cmake
+++ b/riscv64-purecap.cmake
@@ -4,14 +4,14 @@ set(CMAKE_SYSTEM_NAME Linux)
 set(SDK "$ENV{HOME}/cheri/output/sdk" CACHE PATH "path to cheri SDK")
 
 # Set toolchain compilers
-set(CMAKE_C_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-cc)
-set(CMAKE_CXX_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-c++)
+set(CMAKE_C_COMPILER ${SDK}/bin/clang)
+set(CMAKE_CXX_COMPILER ${SDK}/bin/clang++)
 
 # Don't run the linker on compiler check
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 
 # Define sysroot path for CHERI-build`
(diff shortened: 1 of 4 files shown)


alex4o:
bors try

ltratt:
> Now the only linker that can be used is llvm's lld by doing -fuse-ld=lld.

Silly question: why do we want to do this?

alex4o:
Because now cheribuild.py doesn't build a different linker and this is the linker used by cheribsd. I don't know why it isn't set to be the default tho.

ltratt:
Let me rephrase: if I understood your comment we've now "ruled out" other linkers. Is there any reason for us to rule out those other linkers?

alex4o:
I don't know, there might be one, but I need to ask on the CHERI slack, or go through the commits they have made. I can't think of a reason on the top of my head.

riscv64-purecap.cmake
```diff
@@ -4,20 +4,20 @@ set(CMAKE_SYSTEM_NAME Linux)
...
-set(CMAKE_CXX_COMPILER ${SDK}/bin/riscv64-unknown-freebsd13-c++)
+set(CMAKE_C_COMPILER ${SDK}/bin/clang)
+set(CMAKE_CXX_COMPILER ${SDK}/bin/clang++)
 
 # Don't run the linker on compiler check
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 
 # Define sysroot path for CHERI-build`
-set(CMAKE_SYSROOT ${SDK}/sysroot-riscv64-purecap)
+# set(CMAKE_SYSROOT ${SDK}/sysroot-riscv64-purecap)
```
djichthys, line 14: Why is this commented out ?

alex4o: Oh you don't need to set the sysroot manually anymore, it is integrated in the cfg file now.

djichthys:
Do we need to specify the use-ld flag. This will be set once we set the sysroot and is a more comprehensive solution

PR #16, opened 2021-02-24
Added Morello purecap cmake toolchain file

djichthys:
Added Morello pure capability (*purecap*) toolchain file to use with cmake.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -10,11 +10,18 @@ for example in *.c; do
     make bin/"${example%%.*}"
 done
 
-mkdir build 
-cd build
-    
-cmake -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake ..
-make
+# arg-1 : Source directory
+# arg-2 : cmake toolchain file
+function build_cheri_examples()
+{
+    ARCH=$(echo ${2} | cut -d '.' -f 1)
+    BUILD_DIR=${1}/build/${ARCH}
 
-cd ..
-rm -rfv build
+    mkdir -p ${BUILD_DIR}
+    cmake -B ${BUILD_DIR} -S ${1}  -DCMAKE_TOOLCHAIN_FILE=${2}
+    cmake --build ${BUILD_DIR}
+}
+
+build_cheri_examples $(pwd) riscv64-purecap.cmake
+build_cheri_examples $(pwd) morello-purecap.cmake
+rm -rfv $(pwd)/build
--- a/README.md
+++ b/README.md
@@ -1,12 +1,13 @@
 # cheri-examples
-cheri-riscv sample c programs
+cheri sample C programs for RISC-V and Morello targets in *pure capability* (purecap) mode.
 
 ------
 ## Build instructions with CMake
 Build all the example code with CMake using default settings :
 1. Create a separate build directory (\<build-dir\>) outside the source directory (\<source-dir\>).
 2. Generate the build files using either **Unix Makefiles** (*default*) or **Ninja** (specfied
-   with **-G**). Currently only the *RISC-V* cheri-purecap toolchain file has been created.
+   with **-G**). Currently only the *RISC-V* and *Morello* cheri-purecap toolchain files have
+   been created.
    1. If executing from inside the build directory :
 
    `cmake -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake <relative-path-to-source-dir>`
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,12 +1,14 @@
 # cheri-examples
-cheri-riscv sample c programs
+cheri sample C programs for RISC-V and Morello
```
ltratt, line 2: Let's line wrap to the same width as everything else in the file.

ltratt:
Can we/should we link this from `.buildbot.sh`? @alex4o might have thoughts.

alex4o:
Yes we need to, but at the moment I don't build the morello toolchain at all. So I THink we should be able to merge that tommorow.

README.md
djichthys: sorry. Done!

ltratt:
@alex4o Should we wait for that PR before merging this or merge this and then you include the necessary line in `buildbot.sh` in your PR?

djichthys:
No problem with either @alex4o and @ltratt. If we add the other one first, I can add the changes into `.buildbot.sh`  and push more changes through on this PR.

alex4o:
Sorry I havn't explained that very well, I will need to change the buildbot config. The things here wouldn't need changing, so update the .buildbot.sh file. I will do the `bors` r+ things when the new config is ready.

ltratt:
@alex4o What do you mean by "so update the .buildbot.sh file"?

alex4o:
What I mean is that the inclusion of the new test case for building the examples against the morello sdk, should be included in the .buildbot.sh file. And until the pr hasn't been merged everything should be fine.

ltratt:
I'm still confused :) Are you saying that @djichthys needs to do something in this PR or not?

alex4o:
Sry, I am trying very hard to explain what I mean. Yes he needs not to concern himself with anything, as when everything is ready I can just merge the pr, and anything else can just go without having to care about how that happens

ltratt:
OK :) @djichthys Please squash.

djichthys:
HI @ltratt and @alex4o, I have now added instructions to build for both risc-v and morello purecap in `.buildbot.sh` . If this is alright. I will squash push to the repo immediately afterwards. Please review.

.buildbot.sh
```diff
@@ -10,11 +10,19 @@ for example in *.c; do
...
-mkdir build 
-cd build
-    
-cmake -DCMAKE_TOOLCHAIN_FILE=riscv64-purecap.cmake ..
-make
+# arg-1 : Source directory
+# arg-2 : cmake toolchain file
+function build_cheri_examples()
+{
+    CWD=$(pwd)
```
ltratt, line 17: I'm not sure I quite see why we have `$1` (which is `$(pwd)`) and this?

djichthys: Thanks @ltratt . I just thought I could de-couple the source directory from the current working directory within the function.
In this case, the source directory and the current working directory just turned out to be the same, but the function can be reused in other directories. 

If it is over-engineered,easy enough to replace with $1.

README.md
```diff
@@ -1,12 +1,13 @@
...
-cheri-riscv sample c programs
+cheri sample C programs for RISC-V and Morello targets in *pure capability* (purecap) mode.
 
 ------
 ## Build instructions with CMake
 Build all the example code with CMake using default settings :
 1. Create a separate build directory (\<build-dir\>) outside the source directory (\<source-dir\>).
 2. Generate the build files using either **Unix Makefiles** (*default*) or **Ninja** (specfied
-   with **-G**). Currently only the *RISC-V* cheri-purecap toolchain file has been created.
+   with **-G**). Currently only the *RISC-V* and *Morello ARM* cheri-purecap toolchain files have
```
jacobbramley, line 9: I think the proper name would be just `*Morello*`.

.buildbot.sh
cristianurlea: That is only true if one builds within the source tree, otherwise the source directory (`$1`) and the build root (`$(pwd)`) will be different. AFAIK the cmake convention is to always build "out-of-source" which is why there are two flags that one can pass to cmake. This can be seen below on line 22 where the `-B` flag gets the build root path, whereas the `-S` flag gets the source path.

ltratt: Unless I'm reading this incorrectly, `$CWD` is always `==` to `$1`. So I'm not sure we're getting the effect we expect.

djichthys: Changed everything within function to `${1}`.

ltratt:
Please squash.

djichthys:
Squashed and force pushed

PR #17, opened 2021-02-26
Rename the builder used for status for bors.

alex4o:
I have renamed the name of the builder on the buildbot server as the names: cheri-build-script and cheribuild-script are very confusing.
You can see it here: http://bencher14.soft-dev.org:8010/


djichthys:
I think it should be based on the git repo. If we have multiple git repos, the script names should reflect that.

PR #21, opened 2021-06-20
A few operations for keeping the repository in order

drpdr:
- Renamed files to give them a meaningful name
- Deleted redundant code
- Added some comments


ltratt:
Excellent, a first PR is always a good thing :)

set_bounds.c
```diff
@@ -6,25 +6,20 @@
...
 	uint32_t bounds = 64;
-	puts("bounds(64): ");
+
+	puts("Bounds [Choose a value greater than 64]:");
 	if (0 == scanf("%u", &bounds))
 	{
 		error("Extraneous input");
 	}
-
+	// trying to extend the bounds generate an exception
```
ltratt, line 23: Sorry for the nitpicking: we capitalise comments (so "Trying"); and put a full stop at the end. And I think "generate" -> "generates". If you make those changes in a new commit and push them, the best trick we've found is to then add a comment after my comment saying "Fixed in XYZ" where "XYZ" is the commit hash.

drpdr: Fixed in 0ce3570

ltratt:
Please squash.

[That's our term for "I'm happy as the reviewer with this PR. You can now, if you wish, use `git rebase` to reduce the number of commits (while leaving the overall effect of the PR unchanged) and force push an update. You don't have to squash anything, in general, though in this PR you'll definitely want to squash away 0ce3570!]

ltratt:
Oh, and I forgot to say: when something is ready for rereview (a new commit, a squash, etc) please say something like "ready for review" on the PR. Sorry!

drpdr:
Ready for review.

drpdr:
Ready for review.

ltratt:
Ah, you'll need to run clang's formatter on this. That'll produce a diff which you can push as a new commit. I'll review that, then (with luck) we can go through squashing again. If you're not sure if your new commit will work you can issue a `bors try` command which will get bors to try building the PR but won't merge it into master.

drpdr:
Why don't we have something in the CI about this?

ltratt:
We do :) It checks that we've got the formatting right: that's why bors failed above. It's our job to get our commits right -- in theory, every commit should be properly formatted (which is not something a CI could ever sensibly do).

drpdr:
bors try

ltratt:
OK, so now we have some more interesting failures. If you click on `buildbot/capablevms-test-script` you'll see the output from running the build script at the bottom of the page. If you want the complete log, click on "view all 117 lines". I think the problem is that you need to edit at least one `Makefile` in this repo to reflect file name changes (and maybe other bits too).

drpdr:
Right, I forgot the CMakeList!

drpdr:
bors try

ltratt:
Please squash.

drpdr:
Ready for review.

PR #23, opened 2021-06-23
Add Morello support to Makefile

drpdr:
Add Morello support to Makefile

- default architecture is Morello
- ENV files to switch to riscv64

### DIFF
--- a/Makefile.riscv64
+++ b/Makefile.riscv64
@@ -3,7 +3,7 @@ CFORMAT=$(HOME)/cheri/output/sdk/bin/clang-format
 CXX=$(HOME)/cheri/output/sdk/bin/clang++
 CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
 SSHPORT=10021
-export 
+export
 
 cfiles := $(wildcard *.c)
 cppfiles := $(wildcard *.cpp)
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -9,7 +9,7 @@ for example in *.c; do
     if [ "${example}" == "seal.c" ]; then
         continue
     fi
-    make bin/"${example%%.*}"
+    make -f Makefile.riscv64 bin/"${example%%.*}"
 done
 
 # arg-1 : Source directory
--- a/Makefile.morello
+++ b/Makefile.morello
@@ -1,0 +1,31 @@
+CC=$(HOME)/cheri/output/morello-sdk/bin/clang
+CFORMAT=$(HOME)/cheri/output/morello-sdk/bin/clang-format
+CXX=$(HOME)/cheri/output/morello-sdk/bin/clang++
+CFLAGS=-fuse-ld=lld --config cheribsd-morello-purecap.cfg
+SSHPORT=10085
+export
+
+cfiles := $(wildcard *.c)
+cppfiles := $(wildcard *.cpp)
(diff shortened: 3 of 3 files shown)


drpdr:
PR Ready for review.

Makefile
```diff
@@ -1,9 +1,8 @@
-CC=$(HOME)/cheri/output/sdk/bin/clang
-CFORMAT=$(HOME)/cheri/output/sdk/bin/clang-format
-CXX=$(HOME)/cheri/output/sdk/bin/clang++
-CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
-SSHPORT=10021
-export 
+ifndef ARCH
```
ltratt, line 1: I'm pretty sure I'll forget to define `ARCH` and then on risc-v I'll type `make` and be confused at the resulting weird errors! We could either perhaps:

1. use something like `uname` to automatically guess the right architecture
2. have two separate files `Makefile.riscv` and `Makefile.morello` to separate things out in an unambiguous way

I'm fairly neutral between the two.

drpdr: I should SSH into to ``uname`` while cross-compiling. I'll go for the second, then.

ltratt: Should we be cross-compiling on a remote machine? Or should we be saying "actually you need to compile these on a CHERI machine directly"? I would favour the latter (cross compiling always ends in tears!).

alex4o: It is very important to be able to corss-compile imo. As compiling on qemu will take ages.

ltratt: Can we roughly time how long it takes?

Perhaps I've actually over-stated things in one way. I dislike cross-compilation, but I have no problem with making it *possible*. However at the moment I think we *mandate* cross-compilation with the `ssh` stuff?

alex4o: 1. There is no clang compiler on CheriBSD by default. So you need to compile and install it additionally.
2. We do not mandate it. The `ssh` stuff is to run when cross compiling.

drpdr: I didn't mean linking SSH and cross-compiling in my comment above :)

ltratt: OK, I'm making this PR go off-piste, which is bad of me. I think there's a deeper issue here that we can come back to in a different PR.

drpdr:
PR Ready for review.

Makefile.riscv64
```diff
@@ -2,8 +2,8 @@ CC=$(HOME)/cheri/output/sdk/bin/clang
 CFORMAT=$(HOME)/cheri/output/sdk/bin/clang-format
 CXX=$(HOME)/cheri/output/sdk/bin/clang++
 CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
-SSHPORT=10021
-export 
+SSHPORT=10013
```
ltratt, line 5: Is there any reason to change the port number?

drpdr: You are absolutely right. That's my port. I will fix it.

drpdr: Fixed in 56f2d0a.

ltratt:
Please squash.

drpdr:
splat

drpdr:
Right, that's the Makefile in the bot. I'll fix it.

drpdr:
bors try

drpdr:
Fixed in 073b528.

ltratt:
Please squash.

drpdr:
Done.

PR #24, opened 2021-07-02
Add general example regarding bounds and fix typos along the way

drpdr:
- explicitly setting the bounds outside the allowed range
- dereferencing a pointer outside the allowed range
- there were typos in some files and clang_format was missing a flag in the Makefiles


drpdr:
Late small PR (due to technical issues) ready for review whenever you wish.

Makefile.morello
```diff
@@ -2,7 +2,7 @@ CC=$(HOME)/cheri/output/morello-sdk/bin/clang
 CFORMAT=$(HOME)/cheri/output/morello-sdk/bin/clang-format
 CXX=$(HOME)/cheri/output/morello-sdk/bin/clang++
 CFLAGS=-fuse-ld=lld --config cheribsd-morello-purecap.cfg
-SSHPORT=10085
+SSHPORT=12351
```
ltratt, line 5: I don't think this should have changed?

Makefile.morello
```diff
@@ -25,7 +25,7 @@ run-%: bin/% $$(wildcard %.c) $$(wildcard %.cpp)
 	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/$(<F)'
 
 clang-format:
-	$(CFORMAT) $(cfiles)
+	$(CFORMAT) -i $(cfiles)
```
ltratt, line 28: What does `-i` do?

general_bounds.c
```diff
@@ -0,0 +1,45 @@
+#include "include/common.h"
```
ltratt, line 1: Conventionally the local includes come after the "system" includes. [Not a big deal, but we might as well do it!]

general_bounds.c
```diff
@@ -0,0 +1,45 @@
...
+int main()
+{
+	int32_t array[12] = {0};
+	int32_t *typed_array = &array;
+	u_int32_t bounds = 48;
+
+	printf("Explicitly setting the bounds outside the range ");
+	if (strcmp(ARCH, "Morello") == 0)
+	{
+		printf("does not cause any failure on %s. It automatically clears them.\n", ARCH);
```
ltratt, line 26: Automatically clears what?

general_bounds.c
```diff
@@ -0,0 +1,45 @@
...
+	}
+	else
```
ltratt, line 28: I'm not sure this is quite correct: we only know this is true for cheri/risc-v I think?

general_bounds.c
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	uint64_t length = cheri_getlength(typed_array);
+	for (uint32_t counter = 0; counter <= (length / sizeof(int32_t)) + 11; counter++)
+	{
+		inspect_pointer(typed_array + counter);
+		// Read value to crash
+		if (counter == 12)
+		{
+			puts("--> On any architecture dereferencing outside the range"
```
ltratt, line 40: s/puts/printf/ ?

include/common.h
```diff
@@ -5,6 +5,16 @@
 
 #include <stdbool.h>
 
+// Identify the architecture by means of pre-defined macros
+#ifdef __aarch64__
```
ltratt, line 9: I'm not sure that redefining the includes like this is a good idea: it means that people now need to understand `ARCH` and its semantics. I'd suggest just using `#ifdef` in the normal code and getting rid of this define. As a happy bonus that will allow you to change the test above to be something like:

```c
#if defined(__aarch64__)
// do a dereference
#else if defined(__riscv)
// simply increase the bounds
#else
#error Platform not currently supported
#endif
```

In other words, you can change the test from dynamically guessing what should happen to statically enforcing it. That hopefully will also make it clearer to readers.

Makefile.morello
drpdr: It doesn't matter. It's my port. In general, this is pretty random. 12351, instead, is the port I have got assigned by FVP (config didn't work in this case). Anyway, I'll put it back to 10085 (which is a personal mnemonic I don't think anyone cares :))

Makefile.morello
drpdr: "Inplace edit file(s), if specified". Otherwise, I've noticed formatting is not applied.

general_bounds.c
drpdr: All the examples were written following the opposite convention. I've now modified them as you've suggested. I suspect that they were written in that way because ``clang-format`` automatically resets the includes if you don't put a newline to divide them (alphabetical order?).

general_bounds.c
drpdr: There is some code missing here. I'll rephrase too.

general_bounds.c
drpdr: Right.

drpdr:
Fixed in 721eb7d.

drpdr:
PR ready for review.

Makefile.morello
ltratt: I suspect this should be `SSHPORT ?= ...` which allows the user to override it in the environment.

Makefile.morello
ltratt: Ah!

general_bounds.c
ltratt: Oh, if clang-format does that, we should leave them, otherwise we'll just end up fighting it.

general_bounds.c
```diff
@@ -1,45 +1,52 @@
...
 	uint64_t length = cheri_getlength(typed_array);
 	for (uint32_t counter = 0; counter <= (length / sizeof(int32_t)) + 11; counter++)
 	{
 		inspect_pointer(typed_array + counter);
 		// Read value to crash
 		if (counter == 12)
 		{
-			puts("--> On any architecture dereferencing outside the range"
-				 " causes the following exception:");
+			printf("--> On any architecture dereferencing outside the range"
```
ltratt, line 33: I think we can now change this from "on any architecture" to "On Morello" (or should it be "on AArch64"? can we differentiate Morello from other Arm architectures here?).

drpdr: Dereferencing outside the range, as far as I know, it is forbidden on any architecture. Again, dereferencing, not explicitly setting the bounds.

ltratt: Yes, but now this code is guarded by `#defined` it can't be called on anything that isn't AArch64. [Which is why I'm wondering if we can actually restrict it to Morello, since not every AArch64 platform is Morello.]

drpdr: But we did the check on "Aarch64", you are quite right. Well, we wanted an example working in different ways on "riscv64" and "Morello". If we put an ``||`` to allow the same on "riscv6" then the whole point of the general example will be missed, i.e. on "riscv64" it would crash before reaching the other branch.

drpdr: > I think we can now change this from "on any architecture" to "On Morello" (or should it be "on AArch64"? can we differentiate Morello from other Arm architectures here?).
"Morello" is ARMv8, I think that ``clang`` macros allow only to get that "v8".

ltratt: > If we put an || to allow the same on "riscv64" then the whole point of the general example will be missed

That's not what I'm suggesting.

My point is that if you do:

```c
#if defined(aarch64)
// at this point this code can only be on aarch 64
#endif
```

then the `#ifdef` code implicitly excludes other platforms so the "on any architecture" warning is a bit misleading, because that code won't be compiled on anything other than AArch64.

> "Morello" is ARMv8, I think that clang macros allow only to get that "v8".

Then I guess we want something like `#if defined(__aarch64__) and armversion = v8` to capture "Morello"?

drpdr: Right. I'll try to capture that "v8", then.

drpdr:
Fixed in 0542ff7.

drpdr:
PR ready for review.

drpdr:
Fixed in c1fcd68.

drpdr:
PR ready for review.

ltratt:
Please squash.

drpdr:
Done.

ltratt:
Can we update the commit message to bring out the RISC-V vs. Morello issue that led to this PR? Please force push an update.

drpdr:
Done, as per your request.

ltratt:
Can we make clear in the commit message which platform crashes at which point? Please feel free to force push an update.

drpdr:
Please, let me know if this is clear enough or is it too long.

drpdr:
Forgot to reformat something, maybe.

drpdr:
bors try

drpdr:
bors try

drpdr:
OK to squash?

ltratt:
Please squash.

drpdr:
Splat.

PR #25, opened 2021-07-06
Simple test for bounds.c

drpdr:
After some trial/error, simple SH scripts seem the only feasible way to write the tests.
Kyua with or without atf is not feasible.
Even a simple [TAP](http://testanything.org/testing-with-tap/) framework would add
complication to CheriBSD. We said we don't want to add anything more to CheriBSD.
Also, our tests succeed when there is a crash.

This is only one simple test to open a discussion.

### DIFF
--- a/tests/test_bounds.sh
+++ b/tests/test_bounds.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+# This is a simple way to test "bounds.c" by checking ``stderr``.
+set -e
+
+: "${SSHPORT:=10021}"
+if [ -f "../bin/bounds" ]; then
+    scp -P $SSHPORT ../bin/bounds root@127.0.0.1:/root
+    exit_status=0
+    RESULT={{$(ssh -p $SSHPORT -t root@127.0.0.1 'sh -c "/root/bounds 2>&1 > /dev/null"')} && exit_status=1} || true
+    echo -n "bounds... "
+    if [ $exit_status != 0 ]; then
+        echo "FAILED! See below for more details."
+        echo $RESULT
+    else
+        echo "ok"
+    fi
+else
+    echo "Please, build first bounds.c before running the tests."
+    exit 1
+fi


drpdr:
Draft PR ready for discussion, I think.

tests/test_bounds.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# This script is a draft. Note: Bash is not available on CheriBSD.
```
ltratt, line 2: Does this script need bash? Ideally we'd stick to the portable core (I, for one, do not use bash!).

tests/test_bounds.sh
```diff
@@ -0,0 +1,21 @@
...
+# This is a simple way to test the examples by checking ``stderr``.
+
+: "${SSHPORT:=10021}"
```
ltratt, line 5: What does the `:` do here?

tests/test_bounds.sh
```diff
@@ -0,0 +1,21 @@
...
+RESULT_riscv64="fail"
+if [ -f "../bin/bounds" ]; then
+    scp -P $SSHPORT ../bin/bounds root@127.0.0.1:/root
+	RESULT_riscv64=`ssh -p $SSHPORT root@127.0.0.1 -t 'sh -c "/root/bounds 2>&1 > /dev/null"'`
```
ltratt, line 9: Wrong indentation?

tests/test_bounds.sh
drpdr: The comment is to justify some "portable statements" which, in my opinion, are ugly and unreadable compared to Bash 4.x :)

tests/test_bounds.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
ltratt, line 1: We've found (from bitter experience) that it's a good idea to have `set -e` as the first line of shell scripts, as it forces the shell to stop executing as soon as an error is encountered.

tests/test_bounds.sh
drpdr: Assign the default port (the one we have in our repo) in case SSHPORT is not defined in the env.

tests/test_bounds.sh
drpdr: Of course :)

tests/test_bounds.sh
ltratt: I did not know that!

tests/test_bounds.sh
```diff
@@ -0,0 +1,21 @@
...
+    scp -P $SSHPORT ../bin/bounds root@127.0.0.1:/root
+	RESULT_riscv64=`ssh -p $SSHPORT root@127.0.0.1 -t 'sh -c "/root/bounds 2>&1 > /dev/null"'`
+    echo $RESULT_riscv64
+    # note: other forms of equality (test) fail on SH
+    if [ -z "${RESULT_riscv64##*In-address space security exception (core dumped)*}" ] ;then
+        echo "TEST PASSED!"
+    else
+        echo "TEST FAILED!"
+        echo "Output: $RESULT_riscv64"
+fi
```
ltratt, line 17: Wrong indentation.

tests/test_bounds.sh
drpdr: Yes, that's, of course, an excellent choice in general.

tests/test_bounds.sh
cristianurlea: `${parameter:=word}` means set parameter to word if and only if parameter is not already set to a non-empty value. This is great if you have an opinionated $SSHPORT variable already set.

tests/test_bounds.sh
drpdr: That's weird. I can't see even the previous wrong indentation in my editor :|

ltratt:
As @vext01 will tell you, I am allergic to chatty scripts -- I think that the more output we have, the less we read. So I think that for the common "it works case" we should either output nothing or maybe make things look like `cargo test` e.g.:

```
bounds... ok
```

Only if something goes wrong should we output more than that IMHO.

tests/test_bounds.sh
ltratt: There's a mix of tabs and spaces -- I strongly suggest only using spaces to avoid this problem.

drpdr: That's a common problem with VSCode but I thought I had solved it.

tests/test_bounds.sh
drpdr: From the Bash manual (emphasis mine):

> This option applies to the shell environment and each subshell environment separately (see Command Execution Environment), and **may cause subshells to exit before executing all the commands in the subshell**.

After running the SSH command, in fact, it exits.

We have two options here:

- Avoid `set -e`
- executing locally CheriBSD (the redirections could be an issue here?)

ltratt: Are you saying this *definitely* causes a problem or are you saying it *might* cause a problem?

[I use `set -e` with `ssh` extensively and have never observed an issue.]

drpdr: It is a problem actually. After the SSH command, exits. So, it can't run the rest to check the result.

ltratt: I don't understand what "after the SSH command, exits" means.

drpdr: But the problem may be the redirection, I suppose. And we need the redirection for these "raw" tests.

drpdr: > I don't understand what "after the SSH command, exits" means.

You are right. I'll be more clear.

The script runs until this line `RESULT_riscv64=`ssh -p $SSHPORT -t root@127.0.0.1 'sh -c "/root/bounds 2>&1 > /dev/null"'``.

ltratt: I still don't understand.

drpdr: > I still don't understand.

With `set -e`, all these lines are not executed:

`bash 
if [ -z "${RESULT_riscv64##*In-address space security exception (core dumped)*}" ] ;then
        echo "TEST PASSED!"
    else
        echo "TEST FAILED!"
        echo "Output: $RESULT_riscv64"
fi
else
    echo "Please, build first before running the tests."
    exit 1
fi
`

ltratt: Hang on, that line is *supposed* to fail, right? If so you need to do something like `{X && exit 1} || true`.

drpdr: Which line you are referring to?

drpdr: Oh the RESULT! Right.

drpdr:
PR Ready for review. This is now a simple test for bounds.c. We can create other similar tests after merging this PR.

tests/test_bounds.sh
```diff
@@ -1,21 +1,19 @@
 #!/bin/sh
-# This script is a draft. Note: Bash is not available on CheriBSD.
-# This is a simple way to test the examples by checking ``stderr``.
+# This is a simple way to test "bounds.c" by checking ``stderr``.
+set -e
 
 : "${SSHPORT:=10021}"
-RESULT_riscv64="fail"
+RESULT="fail"
```
ltratt, line 6: Do we need to give this variable a default value?

drpdr: Not really. I'll fix that.

tests/test_bounds.sh
```diff
@@ -1,21 +1,19 @@
...
 if [ -f "../bin/bounds" ]; then
     scp -P $SSHPORT ../bin/bounds root@127.0.0.1:/root
-	RESULT_riscv64=`ssh -p $SSHPORT root@127.0.0.1 -t 'sh -c "/root/bounds 2>&1 > /dev/null"'`
-    echo $RESULT_riscv64
-    # note: other forms of equality (test) fail on SH
-    if [ -z "${RESULT_riscv64##*In-address space security exception (core dumped)*}" ] ;then
-        echo "TEST PASSED!"
+    RESULT={$(ssh -p $SSHPORT -t root@127.0.0.1 'sh -c "/root/bounds 2>&1 > /dev/null"') && exit 1} || true
```
ltratt, line 9: There is one confusing thing here: if `bounds` fails, nothing will be printed out. I wonder if we want to do something like this:

```sh
echo -n "bounds... "
<run ssh command>
exit_status=0
{{RESULT=$(ssh -p $SSHPORT -t root@127.0.0.1 'sh -c "/root/bounds 2>&1 > /dev/null"')} && exit_status=1} || true
if [ $exit_status != 0 ]; then
  echo "failed"
  echo $RESULT
else
  echo "passed"
fi
```

(Warning: not tested in any way!)

drpdr: Alright.

drpdr: Hang on, why do you see that "PASSED" there? I've followed the cargo suggestion with "ok".
Maybe, comments (previous) addressed in 020d11d, was necessary before writing "Ready for review"? My bad, sorry.

ltratt: The `exit 1` means that if the command exited with a 0 exit code we don't reach the `echo`s on the following lines.

drpdr: > The `exit 1` means that if the command exited with a 0 exit code we don't reach the `echo`s on the following lines.

Yes, my question was related to the code shown not to your comment. The code in the PR, not your code.

drpdr: Anyway, I'll address and test and also I want to let you know that I have replaced "PASSED" with "OK", so it is less chatty, as you like :) More words in case it fails, instead.

ltratt: > Yes, my question was related to the code shown not to your comment. The code in the PR, not your code.

Yes, I was referring to the code in the PR.

drpdr: > > Yes, my question was related to the code shown not to your comment. The code in the PR, not your code.
> 
> Yes, I was referring to the code in the PR.

But the code in the PR is [this](https://github.com/capablevms/cheri-examples/commit/020d11d9ec60a137b3a1eb70a700cb79526783e7#diff-dee479da72cd420f474a565af7d5db58e9ea75697240cccf3445d0669e7428d0). Again, your comment is correct but the code you commented is outdated. This won't change the meaning of what you said. I am just trying to understand why it is outdated.

ltratt: I promise I am referring to https://github.com/capablevms/cheri-examples/commit/020d11d9ec60a137b3a1eb70a700cb79526783e7#diff-dee479da72cd420f474a565af7d5db58e9ea75697240cccf3445d0669e7428d0 :) Notice that the `exit 1` happens *before* the `echo "bounds... FAILED!`. In other words, if it fails, the user won't be told that.

drpdr:
Fixed in 1d05528.

drpdr:
PR Ready for re-review.

tests/test_bounds.sh
```diff
@@ -3,17 +3,18 @@
...
 
 : "${SSHPORT:=10021}"
-RESULT="fail"
 if [ -f "../bin/bounds" ]; then
     scp -P $SSHPORT ../bin/bounds root@127.0.0.1:/root
     RESULT={$(ssh -p $SSHPORT -t root@127.0.0.1 'sh -c "/root/bounds 2>&1 > /dev/null"') && exit 1} || true
-    if [ -z "${RESULT##*In-address space security exception (core dumped)*}" ]; then
-        echo "bounds... ok"
+    echo -n "bounds... "
+    exit_status=0
```
ltratt, line 10: I don't think this has been tested? This line sets exit_status then the next line immediately reads the same variable, so nothing can have changed inbetween -- and the `ssh` line still has `exit 1` in it.

drpdr: I've just found this. That's a mix of stuff that's why you see two commits. I'll address and push.

drpdr:
Fixed in 22532b8. PR ready for review.

ltratt:
Can I double check that you've tested this (including testing that if you make "bounds" fail (i.e. succeed :))) that the script does the right thing? If so, I think we would do best to squash the excess commits away and merge this PR.

drpdr:
> Can I double check that you've tested this (including testing that if you make "bounds" fail (i.e. succeed :))) that the script does the right thing? If so, I think we would do best to squash the excess commits away and merge this PR.

Tested if either fails or succeed (in our own terms :)). So, I'll squash.

ltratt:
Please squash.

drpdr:
Splat.

drpdr:
NO. STOP, error.

drpdr:
bors try

drpdr:
Now, splat and ready.

PR #26, opened 2021-07-09
Add tests for the examples that rise an "In-address space security exception"

drpdr:
We need only one sh script to test these examples. So, I've removed the previous "test_bounds.sh".


.buildbot.sh
```diff
@@ -2,7 +2,7 @@
 # vim: expandtab sts=0 sw=4 smarttab
 set -e
 
-find . -iname "*.c" -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.hpp" | xargs ~/cheri/output/sdk/bin/clang-format --dry-run -Werror
+#find . -iname "*.c" -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.hpp" | xargs ~/cheri/output/sdk/bin/clang-format --dry-run -Werror
```
ltratt, line 5: Why is this commented out?

tests/run_tests.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# *** NOTE ***
+# The following examples do not need to be
+# tested because they are simply exploring features or
+# they are tests themselves/experiments:
+# allocate.c, check_length.c, check_mask.c, function.c,
+# seal.c, sentry.c, setjmp.c, stackscan.c
+set -e
+
+: "${BUILD_DIR:="../bin"}"
```
ltratt, line 13: Would we ever want this to be something other than `../bin/`?

tests/run_tests.sh
```diff
@@ -0,0 +1,40 @@
...
+: "${SSHPORT:=10021}"
+# Examples that must trigger an "In-address space security exception"
+EXAMPLES="bounds set_bounds general_bounds xor_pointers"
+ARCH=$(ssh -p $SSHPORT -t -o LogLevel=QUIET root@127.0.0.1 uname -m)
+
+for example in ${EXAMPLES}; do
+    if ! [ $(find $BUILD_DIR -prune -empty 2>/dev/null) ]; then
```
ltratt, line 20: I don't feel strongly about this, but conventionally the `!` goes inside the `[...]` -- and I must admit that I'm not entirely sure what the binding of `! [...] && [...]` on the line below might be.

tests/run_tests.sh
```diff
@@ -0,0 +1,40 @@
...
+        if ! [ -z "${ARCH##*riscv*}" ] && [ "$example" = "set_bounds" ]; then
+            echo "Skipping '$example' because you are not on riscv64..."
+            continue
+        fi
+        scp -P $SSHPORT "$BUILD_DIR/$example" root@127.0.0.1:/root
+        exit_status=0
+        # Put the argument, when not needed will be simply ignored
```
ltratt, line 27: What does "put the argument" mean?

ltratt:
In general (and we can experiment with this at squashing time) we prefer to have one commit that does the renaming and then subsequent commits which change things, to try and deal with the fact that git's renaming mechanism doesn't really exist (it works on content similarity, so if you rename + change, the rename is no longer considered a change).

.buildbot.sh
drpdr: No reason. I'll fix it.

tests/run_tests.sh
drpdr: A personal mnemonic that should have been removed.

tests/run_tests.sh
drpdr: > In general (and we can experiment with this at squashing time) we prefer to have one commit that does the renaming and then subsequent commits which change things, to try and deal with the fact that git's renaming mechanism doesn't really exist (it works on content similarity, so if you rename + change, the rename is no longer considered a change).

Are you referring to "test_bounds.sh"?

ltratt: I think this comment ended up in the wrong place -- but yes.

tests/run_tests.sh
drpdr: I have been fiddling with the POSIX core, that separation was helpful to avoid (other) silly errors. This works, but I'll fix the style as you wish.

tests/run_tests.sh
drpdr: > I think this comment ended up in the wrong place -- but yes.

Yes, the place is wrong. So, what's your suggestion?

drpdr: > Would we ever want this to be something other than `../bin/`?

I've added the test script to .buildbot.

ltratt: > Yes, the place is wrong. So, what's your suggestion?

I meant: you replied to a PR-wide comment and put it back on a single line :) It should go in the general "chat" at the bottom of the PR.

ltratt: > >  Would we ever want this to be something other than ../bin/?
>I've added the test script to .buildbot.

I don't see how the two things are related?

ltratt: Oh, because buildbot wants it to be `bin`. OK, makes sense.

drpdr:
> In general (and we can experiment with this at squashing time) we prefer to have one commit that does the renaming and then subsequent commits which change things, to try and deal with the fact that git's renaming mechanism doesn't really exist (it works on content similarity, so if you rename + change, the rename is no longer considered a change).

Double checking if this comment now goes to the right place. What's your suggestion about the renaming/delete?

ltratt:
At squashing time, we can "insert" a new commit before the existing one that just renames the file. Then, I think, your new commit will neatly replay over that new commit. We can't do that without a force push, so it'll have to wait until everything else is done.

drpdr:
> At squashing time, we can "insert" a new commit before the existing one that just renames the file. Then, I think, your new commit will neatly replay over that new commit. We can't do that without a force push, so it'll have to wait until everything else is done.

I may be wrong. But since I've created another file and then removed the previous one, I think I've to find some other option... I was not sure whether I needed to keep test separated until the end.

ltratt:
We can try at squashing time: it might work (or might not).

drpdr:
Fix in f9b4528.

drpdr:
PR ready for review.

ltratt:
Please squash (and try the "insert a commit at the start" thing -- bearing in mind that it might be too difficult!).

drpdr:
I am not sure from where that third commit comes from...

drpdr:
bors try

ltratt:
I think the commits are the wrong way around: the rename commit needs to be the first one.

drpdr:
> I think the commits are the wrong way around: the rename commit needs to be the first one.

I didn't rename. I deleted the existing one at the end because no longer needed. After finishing "run_tests.sh", I deleted the old "test_bounds.sh" because reduntant.

drpdr:
Anyway, let me try to fix .buildbot first, then I can even rethink about this if you don't like it.

drpdr:
bors try

ltratt:
Conceptually, isn't this PR a "rename then change" PR rather than "add then delete" PR? Either way, please fix buildbot first :)

drpdr:
bors try

drpdr:
bors try

drpdr:
@ltratt question: buildbot is failing because it needs to run cheri and qemu in its container. This will add a lot of overhead. Do we need this overhead now or we are happy to have the test and think about this after we have more to test?

ltratt:
Maybe ask @alex4o -- I think it should be semi-lightweight to spin up a CHERI VM from the overnight build on b14? To be honest, it's bound to be the case that we need to run things on an actual CHERI emulator/whatever at some point, so perhaps now is the time to bite the bullet.

alex4o:
currently qemu isn't being build nightly but that can be easily changed. Although startup will be slow no matter what I think It will mostly add around 5/10 min of time to start the vm.

ltratt:
5-10 minutes is fine. @alex4o maybe you can give @drpdr in getting this into buildbot?

drpdr:
bors try

ltratt:
Where are we with this?

drpdr:
Work-in-progress...

tests/run-purecap.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+ARCH=$1
+
+if [ "$ARCH" = "riscv64" ] || [ "$ARCH" = "morello" ];; then
+   NETDEV="user,id=net0,smb=$HOME/cheri<<<source_root@ro:$HOME/cheri/build<<<build_root:$HOME/cheri/output<<<output_root@ro:$HOME/cheri/output/rootfs-$ARCH-purecap<<<rootfs,hostfwd=tcp::10021-:22"
+   $HOME/cheri/output/sdk/bin/qemu-system-"$ARCH"cheri -M virt -m 2048 -nographic -bios bbl-"$ARCH"cheri-virt-fw_jump.bin -kernel $HOME/cheri/output/rootfs-$ARCH-purecap/boot/kernel/kernel -drive if=none,file=$HOME/cheri/output/cheribsd-$ARCH-purecap.img,id=drv,format=raw -device virtio-blk-device,drive=drv -device virtio-net-device,netdev=net0 -netdev $NETDEV -device virtio-rng-pci
```
ltratt, line 9: With buildbot, one thing we have to be careful about is that, IIRC, files dumped in the buildbot user's home directory won't be flushed between runs. In other words, this can create some weird caching effects. It might be worth asking @vext01 about this, as I might be wrong.

drpdr: What if we decide to leave the `worker` with a 1-week old SDK and imgs? And I can add some scheduled tasks to refresh it when needed. In this way, we can always have buildbot running and prevent any future issues.

drpdr: > With buildbot, one thing we have to be careful about is that, IIRC, files dumped in the buildbot user's home directory won't be flushed between runs. In other words, this can create some weird caching effects. It might be worth asking @vext01 about this, as I might be wrong.

Right. So the SDK must be put elsewhere.

ltratt: At least at first, we want to avoid all caching: it'll bite us in weird ways. Later, once we understand what's going on, we can try and be a bit more clever and cache things.

ltratt: > Right. So the SDK must be put elsewhere.

`.buildbot.sh` should download the SDK, build it etc etc. I know it'll be slow, but that doesn't matter for now.

drpdr: > > Right. So the SDK must be put elsewhere.
> 
> `.buildbot.sh` should download the SDK, build it etc etc. I know it'll be slow, but that doesn't matter for now.

OK. I'll put the run inside '.buildbot.sh'. I need to identify an extra target or maybe use the same `nohup` trick with `cheribuild`.

drpdr:
bors try

drpdr:
Wrong PR, my bad.

drpdr:
Ready for a second-pass. This is going to be heavily restructured and cleaned after passing a "try".

drpdr:
bors try

drpdr:
The test-script tries to run the VM but fails because of `samba`. This should be the only issue left before a successful `try`.

drpdr:
bors try

drpdr:
This is now ready for re-review and get back where we were. Some fixes in 488e611.

tests/run_cheri_examples.py
```diff
@@ -0,0 +1,64 @@
...
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+#
```
ltratt, line 36: Let's change these two lines with `#` to just a single blank line (with no `#`).

tests/run_cheri_examples.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+
+from run_tests_common import boot_cheribsd, run_tests_main
+
+
+def run_cheri_examples_tests(qemu: boot_cheribsd.QemuCheriBSDInstance, args: argparse.Namespace) -> bool:
+    if args.sysroot_dir is not None:
+        boot_cheribsd.set_ld_library_path_with_sysroot(qemu)
+    boot_cheribsd.info("Running tests for cheri-examples")
+    # qemu.checked_run("cd '{}'".format(qemu.smb_dirs[0].in_target), timeout=10)
```
ltratt, line 48: Commende code.

tests/run_cheri_examples.py
```diff
@@ -0,0 +1,64 @@
...
+    # qemu.checked_run(args.test_command, timeout=args.test_timeout, pretend_result=0,
+    #                  ignore_cheri_trap=args.ignore_cheri_trap)
+    # This is the BUILD_DIR for the tests, not for this script
+    os.environ["BUILD_DIR"] = "bin"
+    os.system("sh ./tests/run_tests.sh")
```
ltratt, line 53: I think you can remove the line above and then simplify the two lines down into `os.system("BUILD_DIR=bin ./tests/run_tests.sh")`.

tests/run_cheri_examples.py
```diff
@@ -0,0 +1,64 @@
...
+    return True
+
+
+def add_args(parser: argparse.ArgumentParser):
+    parser.add_argument("--ignore-cheri-trap", action="store_true", required=False, default=True,
```
ltratt, line 58: Do we make use of this argument anywhere? I can't see it being used, but perhaps it's important!

tests/run_tests.sh
```diff
@@ -0,0 +1,41 @@
...
+
+for example in ${EXAMPLES}; do
+    if  [ ! $(find $BUILD_DIR -prune -empty 2>/dev/null) ]; then
+        if [ ! -z "${ARCH##*riscv*}" ] && [ "$example" = "set_bounds" ]; then
+            echo "Skipping '$example' because you are not on riscv64..."
+            continue
+        fi
+        scp -o "StrictHostKeyChecking no" -P $SSHPORT "$BUILD_DIR/$example" root@127.0.0.1:/root
+        exit_status=0
+        # RESULT={{$(ssh -p $SSHPORT -t -o LogLevel=QUIET root@127.0.0.1 "/root/$example 68")} && exit_status=1} || true
```
ltratt, line 28: Commented code.

ltratt:
This is looking good! One thing that I can't quite see in the logs is where our "should fail" tests (i.e. those that raise SIGPROT) fail. Can you point those out to me? I'm pretty sure that I'm just missing out on something obvious!

tests/run_tests.sh
drpdr: I am thinking of keeping the "Connections logs" because it is not that ugly and helps to understand what is going on. What do you think? The commented code recalls me that now we log the connections.

ltratt: I don't know what a connection log is :)

drpdr:
> This is looking good! One thing that I can't quite see in the logs is where our "should fail" tests (i.e. those that raise SIGPROT) fail. Can you point those out to me? I'm pretty sure that I'm just missing out on something obvious!

Oh, the "$example... ok", will be "... nok", as you suggested "the cargo way", when they fail. Fail for us mean that they do not crash. If they crash they succeed.

tests/run_cheri_examples.py
drpdr: Yes, we don't need this code.

tests/run_cheri_examples.py
drpdr: I wasn't quite sure about the licence. So, I have copied everything and started with "Inspired by ..." without signing.

ltratt: To be honest, I don't consider the contents of this file to really be a copyright issue w.r.t. the examples in the CHERI repository. I think I would just remove the license entirely.

tests/run_tests.sh
drpdr: "Connected to ... "/"Connection closed" :)

drpdr:
Comments addressed in adccd8e. Ready for review.

.buildbot.sh
```diff
@@ -12,6 +12,33 @@ for example in *.c; do
     make -f Makefile.riscv64 bin/"${example%%.*}"
 done
 
+
```
ltratt, line 15: Unecessary second blank line?

tests/run_cheri_examples.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+# PYTHON_ARGCOMPLETE_OK
```
ltratt, line 2: What does this line mean?

ltratt:
A couple more minor comments, but we're close! Can you make sure you put each individual fix in a separate commit and then put "Fixed in #XYZ" against each comment? It makes it much easier to understand what's been fixed and how -- thanks!

tests/run_cheri_examples.py
drpdr: In `bash` you can use [argcomplete](https://github.com/kislyuk/argcomplete) to have tab completion for `argparse`. It can be quite useful in general. `cheribuild` uses it a lot. However, it is a bit more complicated to make it working in `zsh`. It may be useful to someone else or in the future.

.buildbot.sh
drpdr: Yes.

drpdr:
> A couple more minor comments, but we're close! Can you make sure you put each individual fix in a separate commit and then put "Fixed in #XYZ" against each comment? It makes it much easier to understand what's been fixed and how -- thanks!

Do you mean for the PR in general or for this two fixes you have just requested? This PR is a bit messy so yes, of course, during squashing time there will be some work needed to separate each aspect.

.buildbot.sh
drpdr: Fixed in #64d3842.

tests/run_cheri_examples.py
ltratt: I don't think we should include something specific to one shell here: let's get rid of it.

drpdr: Fixed in 75b33b8.

ltratt:
Please squash.

drpdr:
> Please squash.

Any particular suggestion? :)

ltratt:
Nope, you get to choose :)

drpdr:
Squashed.

ltratt:
That first merge commit shouldn't be here: can we rebase this against upstream/master and get rid of the merge commit?

drpdr:
> That first merge commit shouldn't be here: can we rebase this against upstream/master and get rid of the merge commit?

That was my first squash int his PR which ended up with a third commit. I'll fix that.

drpdr:
Rebased.

PR #27, opened 2021-08-06
Employee example to show permissions in action

drpdr:
Simple example to show how permissions work. A struct is made read-only by setting CHERI_PERM_LOAD.

### DIFF
--- a/employee/Makefile
+++ b/employee/Makefile
@@ -1,0 +1,27 @@
+CC=$(HOME)/cheri/output/sdk/bin/clang
+CFORMAT=$(HOME)/cheri/output/sdk/bin/clang-format
+CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
+SSHPORT=10021
+export
+
+read_only := read_only.c employee.c
+full_privileges := full_privileges.c employee.c
+
+all: bin/full_privileges bin/read_only
+
+bin/read_only: $(read_only)
+	$(CC) $(CFLAGS) $^ -o $@
+
+bin/full_privileges: $(full_privileges)
+	$(CC) $(CFLAGS) $^ -o $@
+
+run:
+	scp -P $(SSHPORT) bin/* root@127.0.0.1:/root
+	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/full_privileges'
+	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/read_only'
+
+clang-format:
+	$(CFORMAT) -i $(cfiles)
+
+clean: 
+	rm -R bin/*
--- a/employee/read_only.c
+++ b/employee/read_only.c
@@ -1,0 +1,42 @@
+/* This program is very similar to "full_privileges.c"
+ * but here we set the permissions of the capability
+ * to LOAD, i.e. read-only. So the user is not allowed to
+ * change the salary.
+ */
+
+#include "../include/common.h"
+#include "include/employee.h"
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int main()
+{
+	struct employee employee = {0, "Good", "Employee", 55000};
+	double salary = 0.0;
+	// set the permissions to read-only
+	struct employee *ro_employee = set_read_only(&employee);
+	inspect_pointer(ro_employee);
+	print_details(ro_employee);
+	printf("The struct is read-only so trying to change the salary will make the program crash...");
+	printf("\nInsert the new salary for this employee: \n");
+	fflush(stdout);
+	scanf("%lf", &salary);
+
+	change_salary(ro_employee, salary);
+	print_details(ro_employee);
+}
+
+/*
+ * Traverse the struct to set the permissions.
+ */
+struct employee *set_read_only(struct employee *e)
+{
+	char *ro_name = (char *)cheri_perms_and(e->name, CHERI_PERM_LOAD);
+	char *ro_surname = (char *)cheri_perms_and(e->surname, CHERI_PERM_LOAD);
+	struct employee *ro_employee =
+		(struct employee *)cheri_perms_and(e, CHERI_PERM_LOAD || CHERI_PERM_LOAD_CAP);
+	ro_employee->name = ro_name;
+	ro_employee->surname = ro_surname;
+	return ro_employee;
+}
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 # Test runner to test the following examples:
-# bounds.c, set_bounds.c, general_bounds.c, xor_pointers.c.
+# bounds.c, set_bounds.c, general_bounds.c, xor_pointers.c,
+# read_only.c (from the employee example)
 #
 # *** NOTE ***
 # The following examples do not need to be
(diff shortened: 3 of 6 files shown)


employee/Makefile
```diff
@@ -0,0 +1,25 @@
...
+CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
+SSHPORT=10021
+export
+
+cfiles := $(wildcard *.c)
+examples := $(patsubst %.c,bin/%,$(cfiles))
+
+all: full_privileges read_only
+
+full_privileges: full_privileges.c employee.c
```
ltratt, line 12: Does this automatically link the two `.o` files that result from compiling the `.c` files?

employee/read_only.c
```diff
@@ -0,0 +1,44 @@
...
+int main()
+{
+	struct employee employee = {0, "Good", "Employee", 55000};
+	double salary = 0.0;
+	// set the permissions to read-only
+	struct employee *ro_employee = set_read_only(&employee);
+	inspect_pointer(ro_employee);
+	print_details(ro_employee);
+	printf("The struct is read-only so trying to change the salary will make the program crash...");
+	fflush(stdout);
```
ltratt, line 22: `printf(); fflush(); printf(); fflush` is equivalent to `printf(); fflush` with a "\n" in the printf string.

employee/read_only.c
```diff
@@ -0,0 +1,44 @@
...
+	scanf("%lf", &salary);
+
+	change_salary(ro_employee, salary);
+	print_details(ro_employee);
+}
+
+/*
+ * Traverse the struct to set the permissions.
+ *
```
ltratt, line 33: Extra line.

employee/read_only.c
```diff
@@ -0,0 +1,44 @@
...
+ */
+struct employee *set_read_only(struct employee *e)
+{
+	char *ro_name = (char *)cheri_perms_and(e->name, CHERI_PERM_LOAD);
```
ltratt, line 37: I *think* that this makes the employee *name* read-only, but it allows me to change the pointer to a name? In other words, the *employee* `struct` isn't immutable, but the things it points to become immutable?

include/common.h
```diff
@@ -1,20 +1,19 @@
-#include <cheri/cheri.h>
-#include <cheri/cheric.h>
+#include <cheriintrin.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 
 void inspect_pointer(void *ptr)
```
ltratt, line 6: What does this function do?

ltratt: Ah, it's really a pretty-printing function. Let's document that. And, in its own commit, we could consider renaming it to something like `print_cheri_permissions(void *ptr)` as that seems a better match.

employee/Makefile
drpdr: Yep

employee/read_only.c
drpdr: Sometimes, cheri does weird thing. It was just part of the debugging. I will remove the extra statements.

employee/read_only.c
drpdr: Sure.

employee/read_only.c
drpdr: If I am not mistaken, if you go a few lines below you should find the `CHERI_LOAD_CAP` on the whole struct. This is the discussion I wanted to start with you.

include/common.h
drpdr: Well, it prints not only the permissions...

employee/Makefile
ltratt: I did not know that!

employee/read_only.c
ltratt: I don't think we can pretend that CHERI can have done anything weird here ;)

include/common.h
ltratt: Still, "inspect" is a pretty vague name :) I know it's not part of this PR, but maybe you can do another PR that renames this function to something more informative? I tend to use https://tratt.net/laurie/src/srep/ for this sort of thing (something like `find . -name "*.[ch]" | xargs srep -m -s inspect_pointer newname` tends to do the trick for me).

employee/read_only.c
ltratt: You're quite right!

ltratt:
Right, let's get fixes to the things that need fixing in and merge this PR. Then we should do another PR, broadly based on this one, which can turn read-only employees back into writeable employees. That's a trickier things to do, but will really help make the example more useful!

ltratt:
Ah, of course, we can't merge this until https://github.com/capablevms/cheri-examples/pull/26 is merged -- we need to add tests for this PR to `.buildbot.sh` once #26 is merged.

include/common.h
drpdr: OK, this will be another PR. Nice tool, BTW :)

ltratt:
@drpdr  we raced apparently posting in the same second!

drpdr:
> @drpdr we raced apparently posting in the same second!

Indeed!

drpdr:
I need the tests and the runner to add some other tests. How can I do this without rebasing @ltratt?

employee/read_only.c
drpdr: Fixed in 0843e79.

employee/read_only.c
drpdr: Fixed in 628c992.

ltratt:
Ah, I see what you mean! Something like `get fetch --all; git merge upstream/master` should do the trick.

drpdr:
> Ah, I see what you mean! Something like `get fetch --all; git merge upstream/master` should do the trick.

Right. And then I'll make the merge commit disappear at the end.

ltratt:
Yes, the (eventual) rebase will solve that problem.

drpdr:
I have added the tests in c14d8f3. This PR is still in a draft state, though.

ltratt:
What else remains to be done?

tests/run_tests.sh
```diff
@@ -25,6 +34,9 @@ for example in ${EXAMPLES}; do
         scp -o "StrictHostKeyChecking no" -P $SSHPORT "$BUILD_DIR/$example" root@127.0.0.1:/root
         exit_status=0
         RESULT={{$(ssh -o "StrictHostKeyChecking no" -p $SSHPORT -t root@127.0.0.1 "/root/$example 68")} && exit_status=1} || true
+        if [ "$example" = "read_only" ]; then
+            example="employee/""$example"
```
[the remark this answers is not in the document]
drpdr: Do you mean the first part ("employee/") or the second (surrounding $example)?

drpdr:
> What else remains to be done?

I think this PR can now move from the draft state.

drpdr:
This PR has moved from *draft* to *ready for review*.

tests/run_tests.sh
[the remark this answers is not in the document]
ltratt: I mean, why not just have `example=employee/$example` rather than all the `""`s?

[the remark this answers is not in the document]
drpdr: > I mean, why not just have `example=employee/$example` rather than all the `""`s?

I can fix that. It is a common practice to "quote everything" in POSIX scripting to avoid unexpected errors--especially for someone like me who forgets everything recently :/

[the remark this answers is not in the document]
ltratt: I've never seen *that* level of quoting before :p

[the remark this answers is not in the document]
drpdr: Fixed in d133748.

ltratt:
Please squash.

drpdr:
Squashed.

PR #29, opened 2021-08-19
Reading from the read-only struct should not cause an In-address Security Exception

drpdr:
Actual tests did not detect this issue because they only test the Exception, NOT where it is caused.
Trying to access more than a member of the read-only struct cause an "In-Address Security Exception" due to:

```
// $HOME/cheri/cheribsd/lib/libc/string/strlen.c:[144-157]
...

#else /* __CHERI_PURE_CAPABILITY__ */
size_t
strlen(const char *str)
{
	const char *p;

	p = str;
// HERE, line 153 accessing *p raises the Exception
	while (*p != '\0')
		p++;

	return (p - str);
}
#endif /* __CHERI_PURE_CAPABILITY__ */
```

This draft PR asks for any pieces of advice or wants to be a mean to discuss the issue. Thanks!

### DIFF
--- a/employee/employee.c
+++ b/employee/employee.c
@@ -15,9 +15,9 @@ void change_salary(struct employee *e, double salary)
 
 void print_details(struct employee *e)
 {
-	printf("Employee Id: %d \n", e->id);
+	printf("Employee Id: %d\n", e->id);
 	printf("Employee Name: %s \n", e->name);
 	printf("Employee Surname: %s\n", e->surname);
-	printf("Employee Salary: %lf\n", e->salary);
+	printf("Employee Salary: %.2lf\n", e->salary);
 	fflush(stdout);
 }
--- a/employee/read_only.c
+++ b/employee/read_only.c
@@ -1,29 +1,31 @@
 /* This program is very similar to "full_privileges.c"
  * but here we set the permissions of the capability
  * to LOAD, i.e. read-only. So the user is not allowed to
- * change the salary.
+ * change, e.g. the salary.
  */
 
 #include "../include/common.h"
 #include "include/employee.h"
+#include <assert.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 
 int main()
 {
-	struct employee employee = {0, "Good", "Employee", 55000};
-	double salary = 0.0;
+	struct employee employee = {1000, "Good", "Employee", 55000.0};
+	printf("--- EMPLOYEE ---\n");
 	// set the permissions to read-only
 	struct employee *ro_employee = set_read_only(&employee);
-	inspect_pointer(ro_employee);
+	assert((cheri_perms_get(ro_employee) & CHERI_PERM_STORE) == 0);
 	print_details(ro_employee);
-	printf("The struct is read-only so trying to change the salary will make the program crash...");
-	printf("\nInsert the new salary for this employee: \n");
+	double new_salary = 65000.0;
+	printf(
+		"\nThe struct is read-only so trying to change the salary to %.2lf will make the program "
+		"crash...\n",
+		new_salary);
 	fflush(stdout);
-	scanf("%lf", &salary);
-
-	change_salary(ro_employee, salary);
+	change_salary(ro_employee, new_salary);
 	print_details(ro_employee);
 }
 
@@ -32,11 +34,7 @@ int main()
  */
 struct employee *set_read_only(struct employee *e)
 {
-	char *ro_name = (char *)cheri_perms_and(e->name, CHERI_PERM_LOAD);
-	char *ro_surname = (char *)cheri_perms_and(e->surname, CHERI_PERM_LOAD);
-	struct employee *ro_employee =
-		(struct employee *)cheri_perms_and(e, CHERI_PERM_LOAD || CHERI_PERM_LOAD_CAP);
-	ro_employee->name = ro_name;
-	ro_employee->surname = ro_surname;
-	return ro_employee;
+	e->name = (char *)cheri_perms_and(e->name, CHERI_PERM_LOAD);
+	e->surname = (char *)cheri_perms_and(e->surname, CHERI_PERM_LOAD);
+	return (struct employee *)cheri_perms_and(e, CHERI_PERM_LOAD | CHERI_PERM_LOAD_CAP);
 }
--- a/employee/Makefile
+++ b/employee/Makefile
@@ -4,6 +4,7 @@ CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
 SSHPORT=10021
 export
 
+employee := full_privileges.c read_only.c employee.c include/employee.h
 read_only := read_only.c employee.c
 full_privileges := full_privileges.c employee.c
 
@@ -23,7 +24,7 @@ run:
 	ssh -p $(SSHPORT) root@127.0.0.1 -t '/root/read_only'
 
 clang-format:
-	$(CFORMAT) -i $(cfiles)
+	$(CFORMAT) -i $(employee)
 
 clean: 
 	rm -R bin/*


ltratt:
The PR message isn't rendering properly: you might need to use triple backquotes.

drpdr:
> The PR message isn't rendering properly: you might need to use triple backquotes.

How is now?

ltratt:
Better :)

employee/employee.c
```diff
@@ -13,9 +13,14 @@ void change_salary(struct employee *e, double salary)
 	e->salary = salary;
 }
 
+/* FIXME:
+ * It crashes after any `char *`, i.e.
```
ltratt, line 17: I can't parse this sentence.

ltratt:
I must admit that I don't understand what problem is being highlighted here: can you try explaining again? Bear in mind that I don't know all the context!

employee/employee.c
drpdr: My bad, sorry. I rephrase. 
One of the following statements, only ONE: 
```
- printf("%s\n", e->name);
- printf("%s\n", e->surname);
- printt("%s\n", e-->alpha_id);
```
Will not cause any crash. If we put more than one of them, like:
```
printf("%s\n", e->name);
printf("%s\n", e->surname);
```
The Exception will occur.

ltratt: I think what you're saying is that `print_details` crashes on the second `printf` (no matter what that `printf` is, provided it references some part of `e`)?

drpdr:
> I must admit that I don't understand what problem is being highlighted here: can you try explaining again? Bear in mind that I don't know all the context!

Sure. So, what we are doing here is setting explicitly the permissions and then showing that we can only read from the struct but we cannot modify it (because it is read-only), right?

employee/employee.c
drpdr: > I think what you're saying is that `print_details` crashes on the second `printf` (no matter what that `printf` is, provided it references some part of `e`)?

Yes. If we try to print more than one element of the `struct`.

ltratt: OK, I have no idea :) @jacobbramley might have a suggestion?

drpdr: This happens after we set the `struct` to *read-only* but we should be able to read, right?

jacobbramley: Each of those `printf`s dereferences two capabilities (`e` and `name`/`surname`/`alpha_id`). Are you sure it's dereferencing `e` that's failing and not, say `surname`?

drpdr: > Each of those `printf`s dereferences two capabilities (`e` and `name`/`surname`/`alpha_id`). Are you sure it's dereferencing `e` that's failing and not, say `surname`?

But they all have the permissions. Or am I missing something obvious?

drpdr: > Each of those `printf`s dereferences two capabilities (`e` and `name`/`surname`/`alpha_id`). Are you sure it's dereferencing `e` that's failing and not, say `surname`?

As you can see it is failing in `strlen.c` so, yes, it is related to one of those `char *`.

ltratt: I think what Jacob is suggesting is: can you cause this crash without `printf` involved? Since you think `strlen` is the culprit, why not make a test that *only* uses `strlen` and that might help you narrow the problem down.

drpdr: > I think what Jacob is suggesting is: can you cause this crash without `printf` involved? Since you think `strlen` is the culprit, why not make a test that _only_ uses `strlen` and that might help you narrow the problem down.

The following statements do not cause any crash:
```
strlen(ro_employee->alpha_id);
strlen(ro_employee->name);
strlen(ro_employee->surname);
```

ltratt: Can you paste a `gdb` backtrace from the crash or is that not possible?

drpdr: > Can you paste a `gdb` backtrace from the crash or is that not possible?

```
Core was generated by `./read_only'.
Program terminated with signal SIGPROT, CHERI protection violation
Capability sealed fault caused by register ca2.
#0  strlen (str=0x40278e90 <printf+64> [rxR,0x40140000-0x407aa000] (sentry) "\217 \001\001[\021\001\002[\200\300\376[\021\001\376#H\021") at /home/cheriworker/cheri/cheribsd/lib/libc/string/strlen.c:153
```

```
(gdb) info register ca2
ca2            0xf11720000babc5060000000040278e90       0x40278e90 <printf+64> [rxR,0x40140000-0x407aa000] (sentry)
```

ltratt: We need the full backtrace: try `bt` at the gdb command line.

drpdr:
```
#0  strlen (str=0x40278e90 <printf+64> [rxR,0x40140000-0x407aa000] (sentry) "\217 \001\001[\021\001\002[\200\300\376[\021\001\376#H\021") at /home/cheriworker/cheri/cheribsd/lib/libc/string/strlen.c:153
#1  0x0000000040280576 in __vfprintf (fp=<optimized out>, locale=<optimized out>, fmt0=<optimized out>, ap=<optimized out>) at /home/cheriworker/cheri/cheribsd/lib/libc/stdio/vfprintf.c:912
#2  0x000000004027eb04 in vfprintf_l (fp=0x403541f0 [rwRW,0x40354020-0x40354590], locale=0x40353620 <__xlocale_global_locale> [rwRW,0x40353620-0x40353780],
    fmt0=0x10075b [rR,0x10075b-0x10076f] "Employee Name: %s \n", ap=0x3fffdffe40 [rwRW,0x3fbfe00000-0x3fffe00000]) at /home/cheriworker/cheri/cheribsd/lib/libc/stdio/vfprintf.c:301
#3  0x0000000040278e90 in printf (fmt=0x40278e90 <printf+64> [rxR,0x40140000-0x407aa000] (sentry) "\217 \001\001[\021\001\002[\200\300\376[\021\001\376#H\021")
    at /home/cheriworker/cheri/cheribsd/lib/libc/stdio/printf.c:57
#4  0x0000000000102808 in print_details (e=0x3fffdffe00 [rR,0x3fffdffe00-0x3fffdffe40]) at employee.c:24
#5  0x00000000001025ea in main () at read_only.c:26
```

ltratt: OK, that doesn't help me, but it might help you and @jacobbramley :)

drpdr: > OK, that doesn't help me, but it might help you and @jacobbramley :)

Can you please check if there is some obvious mistake with the `struct` not related to `CHERI` that I may have done. I don't know, definition, declaration, something I am missing...

employee/read_only.c
```diff
@@ -32,11 +34,13 @@ int main()
...
+	char *ro_alpha_id = (char *)cheri_perms_and(e->alpha_id, CHERI_PERM_LOAD);
+	char *ro_name = (char *)cheri_perms_and(e->name,  CHERI_PERM_LOAD);
 	char *ro_surname = (char *)cheri_perms_and(e->surname, CHERI_PERM_LOAD);
-	struct employee *ro_employee =
-		(struct employee *)cheri_perms_and(e, CHERI_PERM_LOAD || CHERI_PERM_LOAD_CAP);
-	ro_employee->name = ro_name;
-	ro_employee->surname = ro_surname;
-	return ro_employee;
+	struct employee ro_employee = {ro_alpha_id, ro_name, ro_surname, e->salary};
+    struct employee *ro_employee_ptr = &ro_employee;
```
jacobbramley, line 41: This returns a dangling pointer, since `ro_employee` is on the local stack. I suspect that this is causing the strange behaviour.

jacobbramley: We can probably overwrite the original struct, especially since the function is named `set_read_only` and takes a non-`const` pointer to it.

drpdr: If I reassign `e` (passed by ref), so:
`e = (struct employee *) cheri_perms_and(&ro_employee,  CHERI_PERM_LOAD | CHERI_PERM_LOAD_CAP)`

is the same.

jacobbramley: Oh, because `cheri_perms_get(&ro_employee)` should be `cheri_perms_get(ro_employee)`; you're currently testing the permissions of the pointer to the pointer.

drpdr: > Oh, because `cheri_perms_get(&ro_employee)` should be `cheri_perms_get(ro_employee)`; you're currently testing the permissions of the pointer to the pointer.

Damn copy/paste :(

drpdr: Even after removing the dangling pointer in d8e8688, the problem still persists (strlen:153).

ltratt: I think you should assume that the problem is not `strlen`.

drpdr: > I think you should assume that the problem is not `strlen`.

Yes, you are right. But those `char *` must be the problem.

employee/read_only.c
```diff
@@ -32,11 +36,9 @@ int main()
...
+	char *ro_alpha_id = (char *)cheri_perms_and(e->alpha_id, CHERI_PERM_LOAD);
+	char *ro_name = (char *)cheri_perms_and(e->name,  CHERI_PERM_LOAD);
 	char *ro_surname = (char *)cheri_perms_and(e->surname, CHERI_PERM_LOAD);
-	struct employee *ro_employee =
-		(struct employee *)cheri_perms_and(e, CHERI_PERM_LOAD || CHERI_PERM_LOAD_CAP);
-	ro_employee->name = ro_name;
-	ro_employee->surname = ro_surname;
-	return ro_employee;
+	struct employee ro_employee = {ro_alpha_id, ro_name, ro_surname, e->salary};
+	return (struct employee *) cheri_perms_and(&ro_employee,  CHERI_PERM_LOAD | CHERI_PERM_LOAD_CAP);
```
jacobbramley, line 43: If I understand GitHub correctly, there's still a dangling pointer problem here.

drpdr: `&ro_employee`?

jacobbramley:
```
struct employee *set_read_only(struct employee *e)                                                                                                                                                                                                                                                                                                                      
{                                                                                                                                                                                                                                                                                                                                                                       
        e->alpha_id = (char *)cheri_perms_and(e->alpha_id, CHERI_PERM_LOAD);                                                                                                                                                                                                                                                                                            
        e->name = (char *)cheri_perms_and(e->name,  CHERI_PERM_LOAD);                                                                                                                                                                                                                                                                                                   
        e->surname = (char *)cheri_perms_and(e->surname, CHERI_PERM_LOAD);                                                                                                                                                                                                                                                                                              
        return (struct employee *) cheri_perms_and(e,  CHERI_PERM_LOAD | CHERI_PERM_LOAD_CAP);                                                                                                                                                                                                                                                                          
}
```

jacobbramley: `struct employee ro_employee = { ... }` is allocated in the local scope, so taking a pointer to it (`&ro_employee`) and letting it leak outside the scope (by `return`) results in a dangling pointer. This is a use-after-free error, basically, not related to CHERI.

jacobbramley: With that change, it works as it should on Morello.

drpdr: All the locals `char *`. Probably, that's why misbehaves in `strlen.c`. Thanks for your time and effort :)

jacobbramley: I don't think the `char *` variables are relevant here; it's all about the `struct`. In my proposal there are no named locals anyway. I never saw the `strlen` issue but since its input would have come from a dangling pointer, you're firmly into the realm of undefined behaviour.

drpdr: Fixed in 2bc562d. Thank you very much.

drpdr:
This now behaves as expected in 2bc562d. Before leaving draft state, I need to add further tests. Otherwise, merge will fail.

employee/read_only.c
drpdr: > `struct employee ro_employee = { ... }` is allocated in the local scope, so taking a pointer to it (`&ro_employee`) and letting it leak outside the scope (by `return`) results in a dangling pointer. This is a use-after-free error, basically, not related to CHERI.

I knew the use-after-free but I could not think about it here. Thanks for your thorough explanation.

drpdr:
This is now ready for review (34cda65).

ltratt:
Can we also get rid of the `FIXME` now?

drpdr:
> Can we also get rid of the `FIXME` now?

Fixed in 26b1eee.

employee/include/employee.h
```diff
@@ -4,7 +4,7 @@
 
 struct employee
 {
-	int id;
+	char *alpha_id;
```
ltratt, line 7: I think this change is unrelated to this PR? If so, let's put it back to `id` to keep the commit clean.

drpdr: Alright.

drpdr: Fixed in 9079901.

ltratt:
I'm now happy with this. @jacobbramley are you OK with it too?

employee/read_only.c
```diff
@@ -6,23 +6,28 @@
...
 	double salary = 0.0;
+	printf("--- EMPLOYEE ---\n");
 	// set the permissions to read-only
 	struct employee *ro_employee = set_read_only(&employee);
-	inspect_pointer(ro_employee);
+	assert((cheri_perms_get(ro_employee) & CHERI_PERM_STORE) == 0);
 	print_details(ro_employee);
-	printf("The struct is read-only so trying to change the salary will make the program crash...");
+	printf("\nThe struct is read-only so trying to change the salary will make the program "
+		   "crash...\n");
```
jacobbramley, line 24: The indentation looks odd (with mixed tabs and spaces).

employee/read_only.c
```diff
@@ -6,23 +6,28 @@
...
 	printf("\nInsert the new salary for this employee: \n");
 	fflush(stdout);
-	scanf("%lf", &salary);
-
+	if (argc > 1)
+		salary = 65000.0;
```
jacobbramley, line 28: Should this case read the salary from `argv[1]`?

jacobbramley (commented):
All the CHERI stuff looks good. I noticed a couple of nits as I went through it. Sorry for not looking for them sooner!

employee/read_only.c
drpdr: This is done by `clang-format` automatically...

jacobbramley: Odd! Ok, never mind then.

employee/read_only.c
jacobbramley: Well the `scanf` approach used below would make more sense than `atoi`. Really, though, it just looked like a placeholder value.

The example would work fine if it didn't prompt (just picking some arbitrary new value), and then you can remove the argument parsing code entirely. Simplifying boilerplate out of examples is usually a good idea.

drpdr: This looks an excellent suggestion. I'll fix it as you have advised.

drpdr: Fixed in e1b25f1.

jacobbramley:
Looks good to me, thanks!

ltratt:
Please squash.

drpdr:
Splat.

ltratt:
The commit message in d882d9 needs a tweak. The first line is too long (max 80 chars) so github then mangles it. The body of the commit also doesn't quite pinpoint the real problem: the code originally returned a stack pointer to a calling function, which is when Bad Things happened. ["Dangling" covers several possibilities, so we want to be clear about which one happened here.]

Please force push an update.

drpdr:
> the code originally returned a stack pointer to a calling function, which is when Bad Things happened.

5b6c11 should have addressed this.

ltratt:
The first line is still > 80 characters and is a bit obscure. Maybe something like "Fix bug where stack pointer is returned to caller"? Please force push an update.

drpdr:
> The first line is still > 80 characters and is a bit obscure. Maybe something like "Fix bug where stack pointer is returned to caller"? Please force push an update.

Fix in 85f05ae, I think.

ltratt:
Almost there! Now we just need to remove the `-` prefix in the commit body and wrap the commit body to 80 characters (it's currently wrapped too small). One more force push and we should be done I hope!

drpdr:
> Almost there! Now we just need to remove the `-` prefix in the commit body and wrap the commit body to 80 characters (it's currently wrapped too small). One more force push and we should be done I hope!

Done in 1b19b66.

PR #34, opened 2021-08-23
Restore full privileges by explicitly sealing and unsealing

drpdr:
Opening this draft PR to discuss if this is how we want to convey the information to the user.
This is just the first part of the `employee sealed` example. I thought to put all the ways to seal/unseal/restore privileges together at first. However, I think that would be better to have three separate smaller programs. This is the first one: explicitly sealing and unsealing. I am not sure that this is what we want.

### DIFF
--- a/employee/sealed.c
+++ b/employee/sealed.c
@@ -1,0 +1,43 @@
+/*
+ * Starting from a `read_only` struct, this program
+ * restores full privileges using a previous `sealed`
+ * capabilty.
+ */
+
+#include "../include/common.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/sysctl.h>
+
+int main()
+{
+	// Request special capability `sealcap` from the operating system
+	// in order to use it as key to seal `new_small_salary`
+	void *__capability sealcap;
+	size_t sealcap_size = sizeof(sealcap);
+	if (sysctlbyname("security.cheri.sealcap", &sealcap, &sealcap_size, NULL, 0) < 0)
+	{
+		error("Fatal error. Cannot get `securiy.cheri.sealcap`.");
+		exit(1);
+	}
+	assert(cheri_perms_get(sealcap) & CHERI_PERM_SEAL);
+	uint8_t *new_small_salary;
+	// Seal `new_small_salary` using previously requested `sealcap`
+	new_small_salary = (uint8_t *)malloc(sizeof(uint8_t));
+	new_small_salary = (uint8_t *)cheri_seal(new_small_salary, sealcap);
+	assert(cheri_is_sealed(new_small_salary));
+	uint8_t *small_salary;
+	small_salary = (uint8_t *)malloc(sizeof(uint8_t));
+	assert(cheri_perms_get(small_salary) & (CHERI_PERM_LOAD | CHERI_PERM_STORE));
+	pp_cap(small_salary);
+	// Make `small_salary` read-only
+	small_salary = (uint8_t *)cheri_perms_and(&small_salary, CHERI_PERM_LOAD);
+	assert(cheri_perms_get(small_salary) & CHERI_PERM_LOAD);
+	pp_cap(small_salary);
+	// Restore it to read-write using previously sealed capability `new_small_salary`
+	small_salary = (uint8_t *)cheri_unseal(new_small_salary, sealcap);
+	assert(cheri_perms_get(small_salary) & (CHERI_PERM_LOAD | CHERI_PERM_STORE));
+	pp_cap(small_salary);
+}
--- a/employee/Makefile
+++ b/employee/Makefile
@@ -4,7 +4,7 @@ CFLAGS=-fuse-ld=lld --config cheribsd-riscv64-purecap.cfg
 SSHPORT=10021
 export
 
-cfiles := full_privileges.c read_only.c
+cfiles := full_privileges.c read_only.c sealed.c
 employee := $(patsubst %.c,bin/%,$(cfiles))
 
 .PHONY: all run clean


drpdr:
Draft PR ready for discussion.

employee/sealed.c
```diff
@@ -0,0 +1,49 @@
...
+	// in order to use it as key to seal `updated_employee_ptr`
+	sealcap_size = sizeof(sealcap);
+	if (sysctlbyname("security.cheri.sealcap", &sealcap, &sealcap_size, NULL, 0) < 0)
+		error("Something is wrong. Cannot get `securiy.cheri.sealcap`.");
+	assert(cheri_perms_get(sealcap) & CHERI_PERM_SEAL);
+	// Explicitly seal `updated_employee_ptr`
+	sealed_employee = (struct employee *) cheri_seal(updated_employee_ptr, sealcap);
+	assert(cheri_is_sealed(sealed_employee));
+	// Explicitly unseal `updated_employee_ptr`
+	unsealed_employee = (struct employee *) cheri_unseal(sealed_employee, sealcap);
```
cristianurlea, line 42: Consider asserting against `CHERI_PERM_UNSEAL` and/or demo splitting the responsibility of a sealing capability from the reciprocal unsealing capabilities

employee/sealed.c
```diff
@@ -0,0 +1,49 @@
...
+	assert(cheri_is_unsealed(unsealed_employee));
+	ro_employee = unsealed_employee;
```
cristianurlea, line 44: Shouldn't this be assigned to back to `updated_employee` ?
Given that  
 - Dataflow: the ultimate source of `unsealed_employee` is `updated_employee`  not `ro_employee` 
 - Inaccurate semantic name: the result is not read only. 
 - Comment below states this restores read/write permissions which might cause readers to believe this is an exploit.

employee/sealed.c
```diff
@@ -0,0 +1,49 @@
...
+	struct employee employee = {0, "Good", "Employee", 55000};
+	struct employee *ro_employee = set_read_only(&employee);
+	printf("*** `ro_employee` is read-only ***");
+	pp_cap(ro_employee);
+	print_details(ro_employee);
+	// Data to update the employee
+	struct employee updated_employee = {1, "Better", "Unsealed", 250000};
+	struct employee *updated_employee_ptr = &updated_employee;
+	// Pointers for the sealed and unsealed capability
+	struct employee *sealed_employee;
```
cristianurlea, line 27: Consider moving declarations to top or declaring at point of first assignment.

employee/sealed.c
```diff
@@ -0,0 +1,49 @@
...
+	struct employee *unsealed_employee;
+	void *__capability sealcap;
+	size_t sealcap_size;
+
+	// Request special capability `sealcap` from the operating system
+	// in order to use it as key to seal `updated_employee_ptr`
+	sealcap_size = sizeof(sealcap);
+	if (sysctlbyname("security.cheri.sealcap", &sealcap, &sealcap_size, NULL, 0) < 0)
```
cristianurlea, line 35: Consider front-loading this check: If obtaining a sealing capability fails, there's no point in running anything else.

employee/sealed.c
```diff
@@ -0,0 +1,49 @@
...
+	assert(cheri_perms_get(sealcap) & CHERI_PERM_SEAL);
+	// Explicitly seal `updated_employee_ptr`
+	sealed_employee = (struct employee *) cheri_seal(updated_employee_ptr, sealcap);
+	assert(cheri_is_sealed(sealed_employee));
+	// Explicitly unseal `updated_employee_ptr`
+	unsealed_employee = (struct employee *) cheri_unseal(sealed_employee, sealcap);
+	assert(cheri_is_unsealed(unsealed_employee));
+	ro_employee = unsealed_employee;
+	printf("*** Using a previous sealed capability we restore read and write permissions ***\n");
+	printf("*** Note: this operation is completely transparent to the user               ***\n");
```
cristianurlea, line 46: What is meant by transparent to the user here?

employee/sealed.c
drpdr: But the point of the example is to restore `ro_employee` to rw. Otherwise, I don't get it. What are we training to explain here?

employee/sealed.c
drpdr: Oh, sure.

ltratt: We're always in C99-or-later land, so "declare at first assignment" is generally the way to go (though, being C, there are sometimes reasons why that can't be done).

drpdr: I think you both have a good point, being C or C-99. I honestly think I have just followed the example code here :) Anyway, "declare at first assignment" seems reasonable in 2021, I guess.

employee/sealed.c
drpdr: I'll add the assertion, then.

employee/sealed.c
ltratt: I'm not sure the `printf`s are useful, to be honest: people still have to read the code to work out what's going on. I'm generally in favour of programs outputting the bare minimum needed for comprehension.

employee/sealed.c
ltratt: A reasonable question is if, at this point, `ro_employee` is read only or read/write?

employee/sealed.c
drpdr: Yes, I have mixed feelings here. What do you suggest? Removing all the `ptintf`s?

ltratt:
IIUC what this example is trying to show is (roughly):

```c
s = create_sealed_capability();

x = { ... }; // some struct
x = cheri_readonly(x);

x = use_unsealed_capability_to_get_back_rw(x);
```

Have I understood that correctly? If so, maybe for this minimal example we can even just make the `struct` involved a `uint8_t` or something small that doesn't clutter the most important points which are a) getting the sealed capability b) using it to turn read-only -> read-write.

employee/sealed.c
ltratt: If they don't tell us something important, then I would get rid of them.

employee/sealed.c
cristianurlea: And an additional question: what are the contents of the memory pointed to by `ro_employee` at this point ? Are they sourced from the original `ro_employee` or are they sourced from `updated_employee` from which a read-only pointer was never derived?

drpdr: > A reasonable question is if, at this point, `ro_employee` is read only or read/write?

I really struggled to make this readable and understandable. If we want to show that a read-only struct can be "restored" to `rw`, the variable must be the same. We can, of course, rename it since the beginning and give it a (better?) name.

drpdr: > Shouldn't this be assigned to back to `updated_employee` ?
> Given that
> 
> * Dataflow: the ultimate source of `unsealed_employee` is `updated_employee`  not `ro_employee`
> * Inaccurate semantic name: the result is not read only.
> * Comment below states this restores read/write permissions which might cause readers to believe this is an exploit.

It is not an exploit, BUT the whole point of the example is to show "privilege escalation".

ltratt: Not calling it "ro" is probably a good idea!

drpdr:
> IIUC what this example is trying to show is (roughly):
> 
> ```c
> s = create_sealed_capability();
> 
> x = { ... }; // some struct
> x = cheri_readonly(x);
> 
> x = use_unsealed_capability_to_get_back_rw(x);
> ```
> 
> Have I understood that correctly? If so, maybe for this minimal example we can even just make the `struct` involved a `uint8_t` or something small that doesn't clutter the most important points which are a) getting the sealed capability b) using it to turn read-only -> read-write.

OK, good point: avoid the clutter. So, would you suggest this example not being `employee` anymore or just change a small part of the `struct`, e.g. salary? Otherwise, using only a `unit8_t` would be a completely different example.

employee/sealed.c
drpdr: > Not calling it "ro" is probably a good idea!

Alright.

drpdr:
> IIUC what this example is trying to show is (roughly):
> 
> ```c
> s = create_sealed_capability();
> 
> x = { ... }; // some struct
> x = cheri_readonly(x);
> 
> x = use_unsealed_capability_to_get_back_rw(x);
> ```
> 
> Have I understood that correctly? If so, maybe for this minimal example we can even just make the `struct` involved a `uint8_t` or something small that doesn't clutter the most important points which are a) getting the sealed capability b) using it to turn read-only -> read-write.

And Yes you have got it right. This is what we want to convey here.

ltratt:
> Otherwise, using only a unit8_t would be a completely different example.

Why would it be a different example? We can `malloc` it on the heap, get a read-only capability to it, and then use a sealed capability to get a read-write capability. Basically I'm in favour of having the smallest example that shows whatever it is we *really* want to show :)

drpdr:
New smaller and (hopefully) easier to understand version in 21ed87d.

employee/sealed.c
```diff
@@ -0,0 +1,42 @@
...
+	size_t sealcap_size = sizeof(sealcap);
+	if (sysctlbyname("security.cheri.sealcap", &sealcap, &sealcap_size, NULL, 0) < 0){
+		error("Fatal error. Cannot get `securiy.cheri.sealcap`.");
+		exit(1);
+	}
+	assert(cheri_perms_get(sealcap) & CHERI_PERM_SEAL);
+	uint8_t *new_small_salary;
+	// Seal `new_small_salary` using previously requested `sealcap`
+	new_small_salary = (uint8_t *) malloc(sizeof(uint8_t));
+	new_small_salary =  (uint8_t *) cheri_seal(new_small_salary, sealcap);
```
ltratt, line 30: Double space after `=`.

employee/sealed.c
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#include "../include/common.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/sysctl.h>
+
+
```
ltratt, line 14: 2 blank lines might be enough here? :)

employee/sealed.c
```diff
@@ -0,0 +1,42 @@
...
+	assert(cheri_is_sealed(new_small_salary));
+	uint8_t *small_salary;
+	small_salary = (uint8_t *) malloc(sizeof(uint8_t)); 
+	pp_cap(small_salary);
+	// Make `small_salary` read-only
+	small_salary = (uint8_t *) cheri_perms_and(&small_salary, CHERI_PERM_LOAD);
+	assert(cheri_perms_get(small_salary) & CHERI_PERM_LOAD);
+	pp_cap(small_salary);
+	// Restore it to read-write using previously sealed capability `new_small_salary`
+	small_salary = (uint8_t *) cheri_unseal(new_small_salary, sealcap);
```
ltratt, line 40: Don't we need an `assert(...CHERI_PERM_STORE)` here?

employee/sealed.c
drpdr: I believe `clang_format` would have fixed it automatically :) I haven't run it yet, though.

ltratt: Ah. We should always run formatters before pushing new commits, because this sort of thing distracts reviewers otherwise.

drpdr: Yes, you are right. I have a pre-commit hook in my TODO list...

employee/sealed.c
drpdr: Of course, even at the beginning of the program, I think.

employee/sealed.c
drpdr: Fixed in 68148ac.

employee/sealed.c
drpdr: Fixed in 68148ac (automatically).

employee/sealed.c
drpdr: Fixed in c91c2d1.

employee/sealed.c
ltratt: I've never been sure how reliable pre commits are so I run things manually myself. But if a pre commit works for you, that's fine!

ltratt:
Please squash.

drpdr:
Splat.

PR #37, opened 2021-09-03
Morello/makefiles

jacobbramley:
This takes #36 further, separating common platform and target configuration into separate makefiles. It also makes the build commands identical for each example directory, so you don't have to worry about `-I..` and so on.

All examples are covered except for 'example_allocators', which will be addressed later.

Also:
* Fix some problems in individual examples so they will build for morello-purecap and riscv64-purecap.
* Add SPDX headers to new files. These match the project-wide licence, so don't change anything, but I think these are useful, especially for examples that are likely to be copied out of context. I welcome comments on the approach.

~~This does not touch the existing buildbot configuration, nor does it remove the cmake files that buildbot currently depends upon. We should do that later.~~


build/Makefile.vars.common
```diff
@@ -0,0 +1,23 @@
...
+# Set by platform-specific makefiles. For example: ~/cheri/output/sdk
+ifndef SDKBASE
+$(error SDKBASE is not set)
+endif
+
+# For example: morello-purecap
+ifndef PLATFORM
+$(error PLATFORM is not set)
+endif
+
```
ltratt, line 16: Should we add `ifndef SSHPORT` here?

ltratt: Ah, I see that elsewhere we do `ifdef SSHPORT`. It probably would make sense to do all of the checks in one place (presumably here), but perhaps that's for another PR?

ltratt:
Can/should we kill some of the cmake files and at least somewhat port `.buildbot.sh` over to using `make`?

build/Makefile.vars.common
jacobbramley: I decided against it because it allows you to build everything without having a sensible value for `SSHPORT` (e.g. because you don't have a model running).

drpdr:
> Can/should we kill some of the cmake files and at least somewhat port `.buildbot.sh` over to using `make`?

I second this. However, not only `.buildbot.sh` needs to be changed but also `tests/run_tests.sh`. Something `$BUILD_DIR` and `EMPLOYEE_DIR` should consider something like:

```
bin-$ARCH-purecap
```

jacobbramley:
> Can/should we kill some of the cmake files and at least somewhat port `.buildbot.sh` over to using `make`?

Yes! I have another patch that kills the cmake files, and started updating `.buildbot.sh`, but I think @drpdr said he was looking at that so I thought it more useful to submit this PR in its current state. I'd be happy to include it here, though, if that's better.

drpdr:
> > Can/should we kill some of the cmake files and at least somewhat port `.buildbot.sh` over to using `make`?
> 
> Yes! I have another patch that kills the cmake files, and started updating `.buildbot.sh`, but I think @drpdr said he was looking at that so I thought it more useful to submit this PR in its current state. I'd be happy to include it here, though, if that's better.

I was looking into this until realising that I could not modify `.buildbot.sh` before changing the paths and updating the `Makefile`s. Moreover, `.buildbot.sh` calls `tests/run_tests.sh` which currently uses `bin` for both standalone example and `employee`. I anticipate we may not be able to merge this PR without porting `.buildbot.sh` :)

jacobbramley:
>  I anticipate we may not be able to merge this PR without porting `.buildbot.sh` :)

My expected order of operations was:

1. Update makefiles to do what we want to do in the end. (This PR.)
2. Port `.buildbot.sh` to use the new makefiles, rather than cmake.
3. Remove the cmake machinery.

I don't think this PR breaks `.buildbot.sh` as it is because it will continue to use cmake. Am I mistaken?

drpdr:
> > I anticipate we may not be able to merge this PR without porting `.buildbot.sh` :)
> 
> My expected order of operations was:
> 
> 1. Update makefiles to do what we want to do in the end. (This PR.)
> 2. Port `.buildbot.sh` to use the new makefiles, rather than cmake.
> 3. Remove the cmake machinery.
> 
> I don't think this PR breaks `.buildbot.sh` as it is because it will continue to use cmake. Am I mistaken?

`.buildbot.sh` [uses](https://github.com/capablevms/cheri-examples/blob/a0c913818bd83327a1604bbf35c8c4556ad1f43d/.buildbot.sh#L36) makefiles and then run the tests. The cmake call is done at the end of `.buildbot.sh`.

ltratt:
Well, we can sort some thing out by doing...

bors try

ltratt:
OK, so @drpdr wins -- this won't merge as-is :) @jacobbramley if you click on "buildbot/capablevms-test-script", you'll get sent to buildbot's output (99% of the time I want the "stdio" output, and I tend to click on "view XYZ lines" to see all of it).

drpdr:
> OK, so @drpdr wins -- this won't merge as-is :) @jacobbramley if you click on "buildbot/capablevms-test-script", you'll get sent to buildbot's output (99% of the time I want the "stdio" output, and I tend to click on "view XYZ lines" to see all of it).

I didn't want to win, actually :) However, this now fails for a small issue about formatting (first line check in `.buildbot.sh`).

jacobbramley (commented):
I should revert this whole commit. The SDK's clang-format must have been updated at some point, because mine doesn't ask for this change now I've updated it.

ltratt:
If you push a revert commit, we can do a bors try and if it works you can do a force push and squash it away.

jacobbramley:
bors try

ltratt:
> /usr/bin/ld: unrecognised emulation mode: elf64lriscv

That's an odd message. @drpdr might have thoughts?

drpdr:
> > /usr/bin/ld: unrecognised emulation mode: elf64lriscv
> 
> That's an odd message. @drpdr might have thoughts?

I *think* something has changed in `clang13` and we have been affected. They have released their own `docker` images and we haven't run our `.buildbot.sh` since then, IIRC. We have no failed nightly builds, though.

drpdr:
bors try

jacobbramley:
bors try

ltratt:
Where are we with this PR @jacobbramley and @drpdr ?

drpdr:
> Where are we with this PR @jacobbramley and @drpdr ?

I was just checking all the files. Everything seems in order and build properly.

ltratt:
OK @drpdr I'm going to leave you to shepherd this one through :)

drpdr:
OK, @jacobbramley. This does not touch `Makefile`s only but I reckon there were a lot of fixes needed along the way. Please, squash how you see fit.

ltratt:
@jacobbramley When you're done squashing, you'll need to tell @drpdr otherwise he might not notice. I tend to just say "Squashed"!

drpdr:
> @jacobbramley When you're done squashing, you'll need to tell @drpdr otherwise he might not notice. I tend to just say "Squashed"!

Indeed, I had no notifications :) Thanks, @ltratt.

drpdr:
I am no curious if the fallback config works are expected.

bors try

PR #38, opened 2021-09-07
Some minimal code to start a discussion about CInvoke like instructions

drpdr:
As in title. This simple example should demonstrate how to use `ldpblr`, which, among other experimental features, can be used to automatically unseal a pair of capabilities (codecap, datacap).

### DIFF
--- a/employee/employee-cinvoke.c
+++ b/employee/employee-cinvoke.c
@@ -44,7 +44,7 @@ int main()
 	// qemu system crash:
 	// cheri_utils.h:225 "Should only be used with sentry cap"
 	// ~/cheri/morello-qemu/target/cheri-common/cheri_utils.h
-	// A sentry is already sealed (by definition!) and every consequent call to 
+	// A sentry is already sealed (by definition!) and every subsequent call to 
 	// sealing functions will clear the tag!!!
 	
 
--- a/employee/employee-cinvoke.c
+++ b/employee/employee-cinvoke.c
@@ -12,50 +12,39 @@
 #include <stdlib.h>
 #include <sys/sysctl.h>
 
-struct cheri_object
+typedef struct cheri_object
 {
 	void *__capability datacap;
 	void *__capability codecap;
-};
+} CHERI_OBJECT;
 
-void print_salary(uint8_t salary);
-void print_something();
-void invoke(struct cheri_object * pair);
+CHERI_OBJECT * obj;
+
+// TODO: refactor, move this stuff elsewhere
+void print_salary();
+void ldpblr(struct cheri_object * pair);
 void *__capability seal_immediate(void *__capability cap);
-
 
 int main()
 {
-	struct cheri_object * obj = (struct cheri_object *) malloc(sizeof(struct cheri_object)); 
-	uint8_t *small_salary = (uint8_t *) malloc(sizeof(uint8_t));
+	obj = (CHERI_OBJECT *) malloc(sizeof(CHERI_OBJECT)); 
+	uint16_t salary = 1000;
 	pp_cap(&print_salary);
-	pp_cap(small_salary);
-	obj->codecap = &print_something;
-	*small_salary = 12;
-	obj->datacap = small_salary;
+	pp_cap(&salary);
+	obj->codecap = &print_salary;
+	obj->datacap = &salary;
 	assert(!cheri_is_sealed(obj->datacap));
-	// obj->datacap = seal_immediate(small_salary);
 	assert(cheri_is_sentry(obj->codecap));
 	assert(cheri_is_sealed(obj->codecap));
-	// obj->codecap = seal_immediate(&print_salary);
-	invoke(obj);
-	printf("Just in case, this is small_salary: %d\n", *small_salary);
-
-	// qemu system crash:
-	// cheri_utils.h:225 "Should only be used with sentry cap"
-	// ~/cheri/morello-qemu/target/cheri-common/cheri_utils.h
-	// A sentry is already sealed (by definition!) and every subsequent call to 
-	// sealing functions will clear the tag!!!
+	obj = (CHERI_OBJECT *) seal_immediate(obj);
+	assert(cheri_is_sealed(obj));
 	
+	ldpblr(obj);
 
 	return 0;
 }
 
-inline void invoke(struct cheri_object * cheri_obj){
-	// Data for the function
-	// uint8_t * datacap = (uint8_t *) cheri_obj->datacap;
-	// Function pointer (print_salary) of target function
-	// void (*codecap)(void) =  cheri_obj->codecap;
+inline void ldpblr(struct cheri_object * cheri_obj){
 	asm(
 		"ldpblr c0, [%w[pair]]\n\t"
 		: /* output regs */
@@ -63,7 +52,6 @@ inline void invoke(struct cheri_object * cheri_obj){
   		: "c0","c1","c2","c3","c4","c5","c6","c7","c8","c9","c10","c11","c12","c13","c14","c15","c16","c17","c18","clr","d8","d9","d10","d11","d12","d13","d14","d15"
 		  /* clobbered registers */
 	);
-	
 }
 
 inline void *__capability seal_immediate(void *__capability cap){
@@ -77,12 +65,9 @@ inline void *__capability seal_immediate(void *__capability cap){
 	return sealed_cap;
 }
 
-void print_salary(uint8_t salary){
-	printf("Salary: %d\n", salary);
-	fflush(stdout);
+void print_salary(){	
+	uint16_t * salary_ptr = (obj->datacap);
+	printf("Salary: %d\n", *salary_ptr);
 }
 
-void print_something(){
-	printf("Hello, I have been unsealed!?\n");
-}
 
--- a/employee/employee-cinvoke.c
+++ b/employee/employee-cinvoke.c
@@ -52,11 +52,11 @@ int main()
 	
 	obj->datacap = seal_immediate(small_salary);
 	obj->codecap = seal_immediate(&print_salary);
-	printf("After sealing...\n");
-	pp_cap(&print_salary);
-	pp_cap(small_salary);
-	assert(cheri_is_sealed(obj->datacap));
-	assert(cheri_is_sealed(obj->codecap));
+	// printf("After sealing...\n");
+	// pp_cap(&print_salary);
+	// pp_cap(small_salary);
+	// assert(cheri_is_sealed(obj->datacap));
+	// assert(cheri_is_sealed(obj->codecap));
 
 	invoke(obj);
 
@@ -70,10 +70,10 @@ inline void invoke(struct cheri_object * cheri_obj){
 	// void (*codecap)(void) =  cheri_obj->codecap;
 	asm(
 		"ldpblr c0, [%w[pair]]\n\t"
-		"ret\n\t"
 		: /* output regs */
 		: [pair]"r"(cheri_obj) /* input regs */
-  		: /* all caller-saved regs */
+  		: "c0","c1","c2","c3","c4","c5","c6","c7","c8","c9","c10","c11","c12","c13","c14","c15","c16","c17","c18","clr","d8","d9","d10","d11","d12","d13","d14","d15"
+		  /* clobbered registers */
 	);
 	
 }
@@ -81,15 +81,15 @@ inline void invoke(struct cheri_object * cheri_obj){
 inline void *__capability seal_immediate(void *__capability cap){
 	void *__capability sealed_cap;
 	asm(
-		"seal c1, %w[to_seal], rb\n\t" /* 10 encoded form */
-		: [c1] "=r" (sealed_cap)
+		"seal %0, %w[to_seal], rb\n\t" /* 10 encoded form */
+		: "=r" (sealed_cap)
 		: [to_seal] "r" (cap)
-		: /* all caller-saved regs */
+		: 
 	);
 	return sealed_cap;
 }
 
 void print_salary(uint8_t salary){
 	printf("Salary: %d", salary);
-	fflush(stdout);	
+	fflush(stdout);
 }
--- a/employee/employee-cinvoke.c
+++ b/employee/employee-cinvoke.c
@@ -1,8 +1,14 @@
 /* This program  atomically unseal a pair of capabilities
  * (code and data) by calling `ldpblr`, i.e.
  * "Load Pair of capabilities and Branch with Link".
+ * It shows that `ldpblr` can unseal the pair but the unsealed
+ * result is not copied to any intermediate register. Thus,
+ * it is not available outside the call. This program, in fact,
+ * crashes when tries to read `datacap`.
+ * 
  * See "ARM Architecture Reference Manual Supplement Morello"
- * (https://developer.arm.com/documentation/ddi0606/latest).
+ * (https://developer.arm.com/documentation/ddi0606/latest)
+ * for further details regarding `ldpblr`.
  */
 
 #include "../include/common.h"
@@ -20,7 +26,6 @@ typedef struct cheri_object
 
 CHERI_OBJECT * obj;
 
-// TODO: refactor, move this stuff elsewhere
 void print_salary();
 void ldpblr(struct cheri_object * pair);
 CHERI_OBJECT * seal_immediate_pair(CHERI_OBJECT * obj);
@@ -38,7 +43,6 @@ int main()
 	assert(cheri_is_sealed(obj->codecap));
 	obj = seal_immediate_pair(obj);
 	assert(cheri_is_sealed(obj));
-	pp_cap(obj);
 	
 	ldpblr(obj);
 
@@ -47,10 +51,10 @@ int main()
 
 inline void ldpblr(CHERI_OBJECT * obj){
 	asm(
-		"ldpblr c29, [%w[obj]]\n\t"
+		"ldpblr c0, [%w[obj]]\n\t"
 		: /* output regs */
 		: [obj]"r"(obj) /* input regs */
-  		: "c29","c0","c1","c2","c3","c4","c5","c6","c7","c8","c9","c10","c11","c12","c13","c14","c15","c16","c17","c18","clr","d8","d9","d10","d11","d12","d13","d14","d15"
+  		: "c0","c1","c2","c3","c4","c5","c6","c7","c8","c9","c10","c11","c12","c13","c14","c15","c16","c17","c18","clr","d8","d9","d10","d11","d12","d13","d14","d15"
 		  /* clobbered registers */
 	);
 }
@@ -67,8 +71,12 @@ inline CHERI_OBJECT * seal_immediate_pair(CHERI_OBJECT * obj){
 }
 
 void print_salary(){	
+	// Since `obj` is sealed, the following line
+	// causes an `In-address security exception`
 	uint16_t * salary_ptr = (obj->datacap);
 	printf("Salary: %d\n", *salary_ptr);
 }
+
+
 
 
--- a/employee/employee-cinvoke.c
+++ b/employee/employee-cinvoke.c
@@ -21,6 +21,8 @@ struct cheri_object
 void print_salary(uint8_t salary);
 void invoke(struct cheri_object * pair);
 
+// TODO SEAL (immediate) for morello
+
 int main()
 {
 
@@ -49,32 +51,37 @@ int main()
 	// assert(cheri_is_sealed(obj->codecap));
 	// assert(cheri_is_sealed(obj));
 
-	// Seal only its parts, i.e. codecap and datacap
-	obj->datacap = cheri_seal(small_salary, sealcap);
-	obj->codecap = cheri_seal(&print_salary, sealcap);
-	assert(cheri_is_sealed(obj->datacap));
-	assert(cheri_is_sealed(obj->codecap));
+	// FIXME: commented to try without sealing
+	obj->datacap = small_salary;
+	obj->codecap = &print_salary;
+	// Seal only its parts, i.e. datacap and codecap
+	// obj->datacap = cheri_seal(small_salary, sealcap);
+	// obj->codecap = cheri_seal(&print_salary, sealcap);
+	// assert(cheri_is_sealed(obj->datacap));
+	// assert(cheri_is_sealed(obj->codecap));
 
 	// FIXME: "Invalid permission for mapped object"
 	invoke(obj);
 
 	// assuming the above line works should we be able to
-	// print_salary(small_salary)?
+	// print_salary(*small_salary)?
 
 	return 0;
 }
 
 inline void invoke(struct cheri_object * cheri_obj){
-	// Function pointer (print_salary) of target function
-	// void (*codecap)(void) =  cheri_obj->codecap;
 	// Data for the function
 	// uint8_t * datacap = (uint8_t *) cheri_obj->datacap;
+	// Function pointer (print_salary) of target function
+	// void (*codecap)(void) =  cheri_obj->codecap;
 	asm(
 		"ldpblr c0, [%w[pair]]\n\t"
+		"ret\n\t"
 		: /* output regs */
 		: [pair]"r"(cheri_obj) /* input regs */
   		: /* all caller-saved regs */
 	);
+	
 }
 
 void print_salary(uint8_t salary){
--- a/employee/employee-cinvoke.c
+++ b/employee/employee-cinvoke.c
@@ -23,7 +23,7 @@ CHERI_OBJECT * obj;
 // TODO: refactor, move this stuff elsewhere
 void print_salary();
 void ldpblr(struct cheri_object * pair);
-void *__capability seal_immediate(void *__capability cap);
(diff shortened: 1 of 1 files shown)


drpdr:
There is some commented code and FIXME as per its draft state, of course :)

drpdr:
There was an uncommented confusing line. Last version open to discussion is 1e5dab4.

ltratt:
What happens at the moment if you run this? [I don't think answering that question will help me, but @jacobbramley might find it useful :) ]

drpdr:
> What happens at the moment if you run this? [I don't think answering that question will help me, but @jacobbramley might find it useful :) ]

"Invalid permission for mapped object". If, instead, I try to do this from an unsealed object I get a "Cheri SEAL exception". According to the abstraction specification any `cinvoke` instruction should operate on two previously sealed capabilties. Now, for `morello`, as we have said, this may not be the case.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,84 @@
...
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/sysctl.h>
+
+struct cheri_object
+{
+	void *__capability codecap;
+	void *__capability datacap;
```
jacobbramley, line 18: What happens if you swap these over? I'm still looking, but in the Morello pseudo-code it looks like the data capability comes first.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,84 @@
...
+
+	// Request special capability `sealcap` from the operating system
+	// in order to use it as key to seal our capabilities
+	void *__capability sealcap;
+	size_t sealcap_size = sizeof(sealcap);
+	if (sysctlbyname("security.cheri.sealcap", &sealcap, &sealcap_size, NULL, 0) < 0)
+	{
+		error("Fatal error. Cannot get `securiy.cheri.sealcap`.");
+		exit(1);
+	}
```
jacobbramley, line 35: At least on Morello, this isn't actually required, because there's a `SEAL (immediate)` instruction that can seal one of the special CInvoke types without a `sealcap`.

drpdr: aha, 4.4.122 on the Extended ISA, amazing! I'll fix this.

employee/employee-cinvoke.c
drpdr: Right. I'll swap them over.

drpdr: In 51e24fd, after the swap you have suggested, the target is called (so we can reach `print_details`) but it fails with `CHERI protection violation`.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,84 @@
...
+	// uint8_t * datacap = (uint8_t *) cheri_obj->datacap;
+	asm(
+		"ldpblr c0, [%w[pair]]\n\t"
+		: /* output regs */
+		: [pair]"r"(cheri_obj) /* input regs */
+  		: /* all caller-saved regs */
+	);
+}
+
+void print_salary(uint8_t salary){
```
jacobbramley, line 80: We pass ` uint8_t*`, not a `uint8_t`. However, that shouldn't be causing the failure since we don't dereference here.

I would expect a compiler error for the direct call, but not the `invoke(obj)`, since that lacks the type checking of a native C call. (That's an interesting observation by itself, I think, though could probably be neatly solved with C++ templates.)

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,84 @@
...
+	// obj->codecap = &print_salary;
+	// obj->datacap = small_salary;
+	// // Seal `cheri_object` using previously requested `sealcap`
+	// obj = (struct cheri_object *) cheri_seal(obj, sealcap);
+	// assert(cheri_is_sealed(obj->codecap));
+	// assert(cheri_is_sealed(obj));
+
+	// Seal only its parts, i.e. codecap and datacap
+	obj->codecap = cheri_seal(&print_salary, sealcap);
+	obj->datacap = cheri_seal(small_salary, sealcap);
```
jacobbramley, line 53: Also note that `LDPBLR` doesn't seem to automatically unseal `datacap`. This shouldn't matter right now, since we don't actually dereference it, but we will do if you fix my other comment.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,84 @@
...
+		: /* output regs */
+		: [pair]"r"(cheri_obj) /* input regs */
+  		: /* all caller-saved regs */
+	);
+}
+
+void print_salary(uint8_t salary){
+	printf("Salary: %d", salary);
+	fflush(stdout);	
+}
```
jacobbramley, line 83: My guess: Is the standard return mechanism (`ret clr` in purecap) actually able to correctly return from an invocation?

The documentation for Morello's `RET` (4.2.87) indicates that it _can_ automatically unseal the return address, but this is controlled by a system register (configured by the kernel). It's possible that we one of the first teams to actually try using this in CheriBSD on Morello, so that flag (`CCTLR[].SBL`) might not be enabled in the CheriBSD kernel.

Where, exactly, is the exception raised? That will probably answer this question.

drpdr: > Where, exactly, is the exception raised? That will probably answer this question.

Line 80.

jacobbramley: Ah, sorry, I meant which instruction, since I don't trust the source-disassembly mapping :-)

That said, it sounds like it's not the return sequence. I'll think a bit more about possible causes.

employee/employee-cinvoke.c
drpdr: Oh, right. This is a mistake. Yes, it is commented so I could not get a compiler error.

employee/employee-cinvoke.c
drpdr: > Also note that `LDPBLR` doesn't seem to automatically unseal `datacap`. This shouldn't matter right now, since we don't actually dereference it, but we will do if you fix my other comment.

This is another interesting implementation aspect compare to the abstract spec of `cinvoke or like`.

employee/employee-cinvoke.c
drpdr: Tried without sealing and it calls the target and hangs, in 5faa64e.

employee/employee-cinvoke.c
drpdr: In the current state, as per 5faa64e, without sealing anything, it just "hangs" after calling `print_salary`. Why? Is it related to your first question ("My guess...")?

employee/employee-cinvoke.c
drpdr: After using `seal (immediate`, it does not seem to seal anything (the assertions fail). It jumps to the right location and it hangs after printing the `salary`. In f7d6b23.

drpdr:
Interesting behaviour after sealing (immediate). Jump to `print_salary`, hangs, but `cheri_is_sealed` returns false...

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,95 @@
...
+  		: /* all caller-saved regs */
+	);
+	
+}
+
+inline void *__capability seal_immediate(void *__capability cap){
+	void *__capability sealed_cap;
+	asm(
+		"seal c1, %w[to_seal], rb\n\t" /* 10 encoded form */
+		: [c1] "=r" (sealed_cap)
```
jacobbramley, line 85: Why hard-code `c1`? If you let the compiler choose a register (as this does for `to_seal`), it'll probably pick `c0`, which saves a move instruction (since that's used for the return value).

I've forgotten some of the intricacies of the inline assembly syntax but I'm fairly sure that that doesn't actually allocate `sealed_cap` to `c1` anyway, but just assigned it to something named `%[c1]`, so this won't work.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,95 @@
...
+		: [to_seal] "r" (cap)
+		: /* all caller-saved regs */
```
jacobbramley, line 87: That's important (probably) for `ldpblr` because it's calling a function, but this doesn't, so that list really can be empty.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,95 @@
...
+	// Data for the function
+	// uint8_t * datacap = (uint8_t *) cheri_obj->datacap;
+	// Function pointer (print_salary) of target function
+	// void (*codecap)(void) =  cheri_obj->codecap;
+	asm(
+		"ldpblr c0, [%w[pair]]\n\t"
+		"ret\n\t"
+		: /* output regs */
+		: [pair]"r"(cheri_obj) /* input regs */
+  		: /* all caller-saved regs */
```
jacobbramley, line 76: In the example I gave you, I meant for you to expand that list. I don't think there's a handy shortcut for it. The documentation is [here](https://developer.arm.com/documentation/102205/latest) (section 7.1, and for purecap you want AAPCS64-cap), but to save you another detour, we need to mark the following as clobbered (because the invokee might clobber them):
```
c0-c18,clr,d8-d15
```

If that gets even a little complicated, I'd revert to using a plain assembly trampoline:

```
// employee-cinvoke.c
extern void invoke(struct cheri_object * cheri_obj);
...

// employee-cinvoke.s
.align 4
.global invoke
.type invoke, "function"
invoke:
  stp cfp, clr, [csp, #-32]!
  ldpblr c0, [c0]
  ldp cfp, clr, [csp], #32
  ret
```

The compiler will see that as a function call (as it is) and do the right thing. The cost is that it does an extra call, but we can easily improve that later if we need to, and it avoids a potentially-deep rabbit hole.

employee/employee-cinvoke.c
```diff
@@ -0,0 +1,95 @@
...
+}
+
+inline void invoke(struct cheri_object * cheri_obj){
+	// Data for the function
+	// uint8_t * datacap = (uint8_t *) cheri_obj->datacap;
+	// Function pointer (print_salary) of target function
+	// void (*codecap)(void) =  cheri_obj->codecap;
+	asm(
+		"ldpblr c0, [%w[pair]]\n\t"
+		"ret\n\t"
```
jacobbramley, line 73: We shouldn't return here. (Sorry if this is my poor suggestion from an earlier comment.)

This will actually cause an infinite loop, since the `ldpblr` sets `lr` to this address, then this `ret` jumps to itself.

Just delete this instruction, add `clr` to the clobber list (as shown below), and it'll be fine.

employee/employee-cinvoke.c
drpdr: > Why hard-code `c1`? If you let the compiler choose a register (as this does for `to_seal`), it'll probably pick `c0`, which saves a move instruction (since that's used for the return value).

Do you mean `"=r" (sealed_cap)`? This gives me a warning: 
```employee-cinvoke.c:85:11: warning: value size does not match register size specified by the constraint and modifier [-Wasm-operand-widths]
                : "=r" (sealed_cap)```

File: employee/employee-cinvoke.c (hunk 9 above)
Comment (drpdr) [reply employee/employee-cinvoke.c]: After adding the registers to the clobbered list, `qemu` kicks me out:

kicked

Updates In 082b1c0.

Comment (drpdr) [reply employee/employee-cinvoke.c]: Is this related to the fact you mentioned, i.e. the `datacap` is not unsealed/cannot be? Do we need to set the OType somehow or am again wrong and tied to the abstract specs?

File: employee/employee-cinvoke.c (hunk 6 above)
Comment (jacobbramley) [reply employee/employee-cinvoke.c]: No, like this:

```
asm("seal %w[dst], %w[src], rb\n\t"                                                                                                                                                                                                                                                                                                                                   
        : [dst] "=r" (sealed_cap)                                                                                                                                                                                                                                                                                                                                         
        : [src] "r" (cap));
```

Comment (jacobbramley) [reply employee/employee-cinvoke.c]: The named changed only because I tested that to make sure it works, and that's what I picked. The assembly name (in square brackets) can match the C name (in parenthese) if you want, and that's often a good idea for readability.

File: employee/employee-cinvoke.c (hunk 9 above)
Comment (jacobbramley) [reply employee/employee-cinvoke.c]: First, I suggest fixing the problems in the sealing function. I think that will resolve this. With the assembly explicitly writing to `c1`, I think the sealed result will just be dropped (because the compiler doesn't know about it).

In the meantime, I'll try getting `ldpblr` to work with an unsealed capability. If I read the specification correctly, `ldpblr` should accept unsealed inputs, so either I've misunderstood that, or that assertion is wrong.

Comment (jacobbramley) [reply employee/employee-cinvoke.c]: > Is this related to the issue you mentioned, i.e. the `datacap` is not unsealed/cannot be? Do we need to set the OType somehow or am again wrong and tied to the abstract specs?

Sorry, I missed this somehow. Yes, I think it's related, but it's hard to follow the threads now.
"Sealed" and "has a non-zero OType" are the same thing, effectively. You don't need to set it a second time.

Comment (drpdr) [reply employee/employee-cinvoke.c]: > > Is this related to the issue you mentioned, i.e. the `datacap` is not unsealed/cannot be? Do we need to set the OType somehow or am again wrong and tied to the abstract specs?
> 
> Sorry, I missed this somehow. Yes, I think it's related, but it's hard to follow the threads now.

Oh, no, I am sorry. I had to do small modifications comment by comment and link them to specific comments. My mistake.

Comment (drpdr) [reply employee/employee-cinvoke.c]: > In the meantime, I'll try getting `ldpblr` to work with an unsealed capability. 

Done in 2c424aa. It works with unsealed capabilities.

File: employee/employee-cinvoke.c (hunk 6 above)
Comment (drpdr) [reply employee/employee-cinvoke.c]: In 814958b, the capability is sealed and `cheri_is_sealed` is true.

**Comment by drpdr** [comment]
In 814958b, I have sealed only the target since as the specification of `ldpblr` does not say it can unseal the data part too, as @jacobbramley pointed out. However, `qemu` keeps kicking me out. This is so interesting :)

**Comment by drpdr** [comment]
This code on `fvp_morello` causes a `CHERI protection violation` when calling `print_salary`. Is `ldpblr` actually unsealing `codecap`, i.e. `print_datails` as specified?

**Comment by drpdr** [comment]
@ltratt should I keep exploring the `riscv64` part in parallel, maybe I understand the bit I am missing with a different perspective?

**Comment by ltratt** [comment]
Please do whatever you think best.

**Comment by drpdr** [comment]
OK, I think I have a clue. `codecap` and `datacap` should be like "class" and "data". In my case, `datacap` is an argument. Even without sealing it, and assigning a value to `small_salary`, `print_salary` is invoked *but* always prints a "dirty value" of 16. There is no link between the two capabilities, then.

File: employee/employee-cinvoke.c (hunk 9 above)
Comment (drpdr) [reply employee/employee-cinvoke.c]: Saying "it works", it's a mistake, I think. It always print a dirty value of "16". Is using a function with a data argument allowed b the specs? I think `datacap` is supposed to be accessed by the function as part of a "class" (in OO terms) not passed as an argument.

**Comment by drpdr** [comment]
> This code on `fvp_morello` causes a `CHERI protection violation` when calling `print_salary`. Is `ldpblr` actually unsealing `codecap`, i.e. `print_datails`, as specified?

The answer to this is "NO". Any `sentry` is always sealed and any subsequent call to sealing functions, e.g. `seal (immediate), causes the tag to be cleared (and so the capability will be invalid). We cannot `unseal` an invalid capability.

**Comment by drpdr** [comment]
Before redesigning the example I have cleaned up the code a bit to demonstrate what I said about "sentries". In other words, with `seal (immediate)` I was sealing what *cannot* be sealed (because you cannot do it twice!). In 
2972c51.

**Comment by drpdr** [comment]
> Before redesigning the example I have cleaned up the code a bit to demonstrate what I said about "sentries". In other words, with `seal (immediate)` I was sealing what _cannot_ be sealed (because you cannot do it twice!). In
> [2972c51](https://github.com/capablevms/cheri-examples/commit/2972c51f7183af41e3c749a59ca15def4e8fa785).

Now, cfe6696.

**Comment by drpdr** [comment]
> OK, I think I have a clue. `codecap` and `datacap` should be like "class" and "data". In my case, `datacap` is an argument. Even without sealing anything, and assigning a value to `small_salary`, `print_salary` is invoked _but_ always prints a "dirty value" of 16. There is no link between the two capabilities, then. I believe I have a _design_ issue in this example.

@jacobbramley do you agree with me on this point? We cannot use an argument as `datacap`. In other words, `print_salary` is wrong in this case.

**Comment by drpdr** [comment]
@jacobbramley this is supposed to work, right? In bd4f5aa.

File: employee/employee-cinvoke.c [hunk 10]
```diff
@@ -0,0 +1,73 @@
...
+		: [pair]"r"(cheri_obj) /* input regs */
+  		: "c0","c1","c2","c3","c4","c5","c6","c7","c8","c9","c10","c11","c12","c13","c14","c15","c16","c17","c18","clr","d8","d9","d10","d11","d12","d13","d14","d15"
+		  /* clobbered registers */
+	);
+}
+
+inline void *__capability seal_immediate(void *__capability cap){
+	void *__capability sealed_cap;
+	asm(
+		"seal %[sealed_cap], %[cap], rb\n\t"
```
jacobbramley, line 60: Shouldn't it be `lpb`, not `rb`?

drpdr: Yes! You are quite right. That tiny argument in a single string (and without different syntax highlighting) you even forget the existence of 🙂

drpdr: Fixed and renamed in 3aad2c3. Still `ldpblr` does not seal it: "CHERI protection violation. Capability sealed fault".

drpdr:
> Before [e83f70e](https://github.com/capablevms/cheri-examples/commit/e83f70e1fbf75e754ec9310bedd1cf45fe40c08c), I tried and debugged all the combinations of _registers_ clobbering/not clobbering `c0` and `c29`---just in case I was not missing anything obvious. I always have the same violation but this time inside `print_salary`, i.e. line 70. So, `ldpblr` is able to unseal `obj`, call the target but when the target is called finds `obj` sealed (?), @jacobbramley.

Unless this is the designed behaviour, @jacobbramley. In this case, we should keep this example as one causing a (non-obvious) *In-address security exception* (Cheri protection sealed fault).

drpdr:
Any "external and fresh" thought about this, @ltratt?

ltratt:
Let's wait for you and @jacobbramley to come to a fixed point, then you can tell me what I might want to look at.

employee/employee-cinvoke.c
jacobbramley: The example will work if `obj` is not sealed. I'm not sure if that grants useful security properties for us, but it's something to start from, at least, and we should probably keep it as a "basic use" example.

As mentioned elsewhere, `ldpblr` will only unseal `obj` if `%[sealed_cap]` is `c29`. I'd overlooked that at the start of this, but it's actually quite important for our security properties. However, demonstrating that is tricky because `c29` is the frame pointer, which might confuse our C toolchain.

I think I would leave the unsealing `c29` variant as an extension task. Doing it neatly probably requires some more assembly, and makes a more complicated example.

jacobbramley:
> Before [e83f70e](https://github.com/capablevms/cheri-examples/commit/e83f70e1fbf75e754ec9310bedd1cf45fe40c08c), I tried and debugged all the combinations of _registers_ clobbering/not clobbering `c0` and `c29`---just in case I was not missing anything obvious. I always have the same violation but this time inside `print_salary`, i.e. line 70. So, `ldpblr` is able to unseal `obj`, call the target but when the target is called finds `obj` sealed (?), @jacobbramley.

Right. `ldpblr` unseals `obj` but only uses the unsealed form as an intermediate value. It is not stored back to any register, so you cannot inspect the unsealed form.

Regarding clobbers: put simply, you need to list anything that your assembly writes, except for parameterised inputs and outputs. Because this is a call, that means all callee-saved registers (c0-etc). The `obj`-unsealing version, with `c29`, requires that you list `c29` as a clobber because it is not an input or output _of the asm block_.

employee/employee-cinvoke.c
drpdr: > As mentioned elsewhere, `ldpblr` will only unseal `obj` if `%[sealed_cap]` is `c29`. I'd overlooked that at the start of this, but it's actually quite important for our security properties. However, demonstrating that is tricky because `c29` is the frame pointer, which might confuse our C toolchain.

Oh, you meant `c29` also in `seal (immediate)` not only as destination in `ldpblr`.

drpdr: > The example will work if `obj` is not sealed. I'm not sure if that grants useful security properties for us, but it's something to start from, at least, and we should probably keep it as a "basic use" example.

This example does not work now demonstrating what we CANNOT do, which I think is not that obvious. It works, instead when `obj` is not sealed. @ltratt what do you think? What would you like to have to finish this PR?

jacobbramley: > > As mentioned elsewhere, `ldpblr` will only unseal `obj` if `%[sealed_cap]` is `c29`. I'd overlooked that at the start of this, but it's actually quite important for our security properties. However, demonstrating that is tricky because `c29` is the frame pointer, which might confuse our C toolchain.
> 
> Oh, you meant `c29` also in `seal (immediate)` not only as destination in `ldpblr`.

No, just in `ldpblr`, sorry. I may have confused things.

drpdr:
OK, @ltratt. TLDR; We have two options here:

1) Showing `ldpblr` without sealing (simple minimal use of `lpdblr`), as @jacobbramley has suggested; 
2) current state, i.e. showing what we CANNOT do with a simple `lpdblr` when sealing `obj`.

Of course, 1) and 2) can be two separate PRs and you can decide which point you would like to address in this PR (exiting draft state).

ltratt:
I need a simpler summary :) Why do I care about `LDPBLR`? Is this PR showing that we can't unseal things or ... ?

drpdr:
> I need a simpler summary :) Why do I care about `LDPBLR`? Is this PR showing that we can't unseal things or ... ?`

`LDPBLR` it is the `cinvoke-like` instruction in `morello`. This example, in the current state, shows that even if `ldpblr` can unseal a `struct` containing pair of capabilities (`datacap`, `codecap`), when it executes the target (`codecap`), the latter cannot unseal the pair. This is because "ldpblr unseals obj but only uses the unsealed form as an intermediate value. It is not stored back to any register, so you cannot inspect the unsealed form." (@jacobbramley a few comments above).

Another possibility is to avoid *sealing* the pair and let `ldpblr` *branch and link* showing how it executes the target.

We acknowledged we learnt how this instruction behaves thanks to this messy PR and discussion :)

drpdr:
@ltratt before exiting draft state, check the comment at the beginning of the file in a163319 and let's see if I can convince you :)

ltratt:
So my question is: what does the example show at a high-level? If I'm being blunt, all I can currently see is that it shows that `ldpblr` does what it says it does at a low-level, and I'm not sure if that's useful to readers or not.

drpdr:
> So my question is: what does the example show at a high-level? If I'm being blunt, all I can currently see is that it shows that `ldpblr` does what it says it does at a low-level, and I'm not sure if that's useful to readers or not.

First, there are no examples showing how to use `ldpblr` and `seal (immediate)`. Second, I *think* that not finding the `pair` unsealed after the call is not obvious, i.e. I would have expected to `obj` being unsealed. This is an extra security implementation detail `morello`. This is `morello` only and so it requires an `#ifdef`, of course.

ltratt:
I think, then, what you're saying is that you didn't expect `ldpblr` to only temporarily unseal? If I look at the docs they say:

>Load Pair of capabilities and Branch with Link calculates an address from the source Capability register, loads from memory two capabilities, a target capability and a data capability. The instruction writes the data capability to the destination Capability register and branches to the target capability, setting C30 to PCC+4.

Is your point that "writes the data capability to the destination Capability register" is misleading?

drpdr:
> I think, then, what you're saying is that you didn't expect `ldpblr` to only temporarily unseal? If I look at the docs they say:
> 
> > Load Pair of capabilities and Branch with Link calculates an address from the source Capability register, loads from memory two capabilities, a target capability and a data capability. The instruction writes the data capability to the destination Capability register and branches to the target capability, setting C30 to PCC+4.
> 
> Is your point that "writes the data capability to the destination Capability register" is misleading?

Well, this should work if we use `c29` but, as we discussed earlier in this PR, `c29` is the `CFP` so this confuses the toolchain. However, you may be right and my interpretation may not be shared by others. So, apart of "how to use lp* and seal" this PR does not show anything else :/

ltratt:
I'm still confused so on that basis I'm going to propose that we close this PR -- I am not sure that struggling to understand what it's showing is doing anything useful and it's sucking up our collective time. On the plus side, this PR has helped (if indirectly) move the conversation forward, so arguably it's served its purpose to us :)

drpdr:
> I'm still confused so on that basis I'm going to propose that we close this PR -- I am not sure that struggling to understand what it's showing is doing anything useful and it's sucking up our collective time. On the plus side, this PR has helped (if indirectly) move the conversation forward, so arguably it's served its purpose to us :)

Alright. This PR has helped (indirectly) to move the conversation forward, as you said. We can close it, then.

ltratt:
OK. I'll leave you to close it (the first time you close a PR it hurts... but you do get used to it! I've closed quite a lot of my own PRs over time).

PR #39, opened 2021-09-21
Some simple interactions with the Default Data Capability register

drpdr:
This PR shows some simple interactions with the DDC:

- basic_ddc.c: very basic compartment switching
- ddc_null.c: crash when writing NULL
- ddc_invalid.c: crash when writing an invalid capability

### DIFF
--- a/hybrid/Makefile.common
+++ b/hybrid/Makefile.common
@@ -1,0 +1,50 @@
+# Set by platform-specific makefiles. For example: ~/cheri/output/sdk
+ifndef SDKBASE
+$(error SDKBASE is not set)
+endif
+
+# For example: morello-purecap
+ifndef PLATFORM
+$(error PLATFORM is not set)
+endif
+
+CC := $(SDKBASE)/bin/clang
+CFORMAT := $(SDKBASE)/bin/clang-format
+
+RUNUSER ?= root
+RUNHOST ?= localhost
+RUNDIR ?=
+CFLAGS ?=
+
+BINDIR := bin/$(PLATFORM)
+CFILES := $(wildcard *.c)
+HFILES := $(wildcard include/*.h)
+BINS := $(patsubst %.c,$(BINDIR)/%,$(CFILES))
+RUNTGTS := $(patsubst $(BINDIR)/%,run-%,$(BINS))
+
+.PHONY: all clang-format clean $(RUNTGTS)
+
+all: $(BINS)
+
+# Remove outputs, but only remove the top-level 'bin/' if it's empty. This
+# mitigates the risk of accidental damage, e.g. if run from a user's home when
+# they have a ~/bin.
+clean:
+	rm -rf $(BINDIR)
+	rmdir bin --ignore-fail-on-non-empty
+
+clang-format:
+	$(CFORMAT) -i $(CFILES) $(HFILES)
+
+$(BINDIR)/%: %.c $(HFILES)
+	@mkdir -p $(BINDIR)
+	$(CC) $(CFLAGS) $< -o $@
+
+$(RUNTGTS): run-%: $(BINDIR)/%
+ifdef SSHPORT
+	scp -P $(SSHPORT) $^ $(RUNUSER)@$(RUNHOST):$(RUNDIR)
+	ssh -p $(SSHPORT) $(RUNUSER)@$(RUNHOST) -t ./$(<F)
+else
+	@echo "'$@' requires SSHPORT to be defined."
+	@false
+endif
--- a/hybrid/basic_ddc.c
+++ b/hybrid/basic_ddc.c
@@ -1,0 +1,31 @@
+/***
+ * This program shows a very basic level of compartmentalisation. It first
+ * `malloc` two blocks, and then flip the DDC to point from one to the other.
+ ***/
+
+#include "../include/common.h"
+#include "include/utils.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+int main()
+{
+	uint16_t *some_int_ptr = (uint16_t *) malloc(sizeof(uint16_t));
+	*some_int_ptr = 200;
+	assert(cheri_address_get(cheri_ddc_get()) != some_int_ptr);
+	uint32_t *some_other_int_ptr = (uint32_t *) malloc(sizeof(uint32_t));
+	write_ddc((void *__capability) some_other_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) != some_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) == some_other_int_ptr);
+	// Note: this program is very simple and writing to the DDC in this fashion
+	// would cause a crash if the program were to execute much further.
+	write_ddc((void *__capability) some_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) != some_other_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) == some_int_ptr);
+	return 0;
+}
--- a/hybrid/ddc_invalid.c
+++ b/hybrid/ddc_invalid.c
@@ -1,0 +1,23 @@
+/***
+ * This program causes an "In-address space security Exception" by clearing the
+ * tag of the capability in the DDC
+ ***/
+
+#include "../include/common.h"
+#include "include/utils.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+int main()
+{
+	// Before clearing the tag we ensure we have a valid one
+	assert(cheri_tag_get(cheri_ddc_get()));
+	// Clearing the tag will cause the exception
+	write_ddc(cheri_tag_clear(cheri_ddc_get()));
+	return 0;
+}
--- a/hybrid/ddc_null.c
+++ b/hybrid/ddc_null.c
@@ -1,0 +1,23 @@
+/***
+ * This program shows that putting "NULL" into the DDC causes an
+ * "In-address space security exception"
+ ***/
+
+#include "../include/common.h"
+#include "include/utils.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+int main()
+{
+	// Ensure the DDC contains a valid capability
+	assert(cheri_tag_get(cheri_ddc_get()));
+	// Putting a NULL will cause the exception
+	write_ddc(NULL);
+	return 0;
+}
--- a/hybrid/include/utils.h
+++ b/hybrid/include/utils.h
@@ -1,0 +1,25 @@
+/***
+ * Inline Assembly utility functions to read/write to/from the Default Data
+ * Capability Register (DDC)
+ ***/
+
+#if !defined(__aarch64__)
+#error "This utility functions are Morello only"
+#endif
+
+void write_ddc(void *__capability cap);
+void *__capability read_ddc();
+
+// Write a capability to the DDC
+inline void write_ddc(void *__capability cap)
+{
+	asm("MSR DDC, %[cap]\n\t" : : [cap] "C"(cap) : "memory");
+}
+
+// Read a capability from the DDC
+inline void *__capability read_ddc()
+{
+	void *__capability ddc_cap;
+	asm("MRS %[ddc_cap], DDC\n\t" : [ddc_cap] "=C"(ddc_cap) : :);
+	return ddc_cap;
+}
--- a/hybrid/Makefile.morello-hybrid
+++ b/hybrid/Makefile.morello-hybrid
@@ -1,0 +1,7 @@
+CHERIBASE ?= $(HOME)/cheri
+SDKBASE ?= $(CHERIBASE)/output/morello-sdk
+CFLAGS := --config cheribsd-morello-hybrid.cfg $(CFLAGS)
+PLATFORM := morello-hybrid
+export
+
+include Makefile.common
--- a/include/common.h
+++ b/include/common.h
@@ -7,7 +7,7 @@
  * Print information about a capability
  */
 
-void pp_cap(void *ptr)
+void pp_cap(void *__capability ptr)
 {
 	uint64_t length = cheri_length_get(ptr);
 	uint64_t address = cheri_address_get(ptr);


drpdr:
Draft PR because I am not sure what you'd expect from such a register (DDC).

hybrid/Makefile.common
```diff
@@ -0,0 +1,50 @@
...
+
+.PHONY: all clang-format clean $(RUNTGTS)
+
+all: $(BINS)
+
+# Remove outputs, but only remove the top-level 'bin/' if it's empty. This
+# mitigates the risk of accidental damage, e.g. if run from a user's home when
+# they have a ~/bin.
+clean:
+	rm -rf $(BINDIR)
```
ltratt, line 33: This removes `BINDIR` so the comment above is redundant (also `rmdir --ignore-fail-on-non-empty` isn't portable, so we don't want to use that).

drpdr: This is from a past PR, copy/pasted and already approved as correct :)

hybrid/ddc_bounds.c
```diff
@@ -0,0 +1,21 @@
+#include "../include/common.h"
+#include "include/utils.h"
+#include <stdint.h>
+#include <stdlib.h>
+#if !defined(__aarch64__)
+#error "This example is Morello only"
+#endif
```
jacobbramley, line 7: I'd put that on the implementation of `write_ddc`, since that's the only Morello-specific part (and could be adapted for other CHERI implementations, potentially).

hybrid/Makefile.common
ltratt: Alas, copy and paste does not prove anything is correct -- you'll need to prove to me that this is correct again (because I'm rather certain that the comment is completely incorrect, unless `rm -rf` magically doesn't delete a directory!).

ltratt:
Can we add a bit more description to the `.c` files saying what each is supposed to do? At the moment the description in the PR is too terse for me to really work out what's supposed to happen. Thanks!

hybrid/Makefile.common
jacobbramley: It was from my first makefiles PR (#36), but the latest one (#37) does a slightly better job.

hybrid/ddc_bounds.c
```diff
@@ -0,0 +1,21 @@
...
+
+int main(){
+    int32_t array[16] = {0};
+	int32_t *array_ptr = &array;
```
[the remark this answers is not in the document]
drpdr: Super right, not `clang-formatted`. This PR would not pass in many ways, especially because we are not *hybrid-ready* yet with *buildbot*.

hybrid/ddc_bounds.c
```diff
@@ -0,0 +1,21 @@
...
+    pp_cap(cheri_ddc_get());
+    int32_t *__capability array_cap = (int32_t *__capability) array_ptr;
```
jacobbramley, line 13: That casts an `int32_t*` to an `int32_t* __capability`, which I think just derives the metadata from the DDC. That means that this first bit just sets the address of the DDC. Am I mistaken? I don't think that derivation sets the bounds.

hybrid/ddc_bounds.c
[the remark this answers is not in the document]
drpdr: OK.

hybrid/Makefile.common
drpdr: But that was not my PR, so I assumed was the preferred/approved way for a `Makefile.common`.

jacobbramley: @ltratt It's not perfect. It will, of course, delete your `./bin/morello-purecap` (or similar) directory, but that's rather less likely to be hit by accident than the `./bin` that we used to clean.

ltratt: Oh, hang on: `BINDIR` is a *subdir* of `bin/`. Hmm. That is a bit confusing! Perhaps it's not worth obsessing over, though. Let's not worry about this one anymore.

drpdr:
> Can we add a bit more description to the `.c` files saying what each is supposed to do? At the moment the description in the PR is too terse for me to really work out what's supposed to happen. Thanks!

More comments added in 31698ba.

hybrid/ddc_bounds.c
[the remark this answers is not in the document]
drpdr: Clang-format fixed in dc16ecf.

hybrid/ddc_bounds.c
drpdr: I still have that `dcc_get` issue so I have to also read from the DDC using `read_ddc` (*morello only*).

hybrid/ddc_bounds.c
drpdr: So, casting here is not sufficient...

hybrid/ddc_bounds.c
```diff
@@ -0,0 +1,21 @@
+#include "../include/common.h"
```
ltratt, line 1: I *think* this file (which is missing a comment?) should be called `ddc_out_of_bounds`?

drpdr: Sure. It does not have a comment yet because I wanted to address @jacobbramley's comment first. It works but I am not sure if it is the intended behaviour.

drpdr:
> @drpdr Can we create (or maybe better, adjust the existing?) "tests" that `assert` that things will be one way or the other rather than just printing values out? For example (in pseudocode) I'd like to see a test something like:
> 
> ```
> x = malloc(y);
> set_ddc_to(<something that doesn't cover x..x+y>);
> assert(capability_doesnt_contain_address(get_ddc(), x));
> ```
> 
> The advantage of `assert` tests is that they integrate well with CI. The disadvantage is that they don't test that a program actually crashes when it does something naughty, so we also need some "this actually SIGPROTs" tests too.
> 
> Another test I'd like to see is something like:
> 
> ```
> x = malloc(y);
> assert(capability_does_contain_address(get_ddc(), x));
> a = malloc(b);
> set_ddc_to(<a..a+b>);
> assert(capability_does_contain_address(get_ddc(), a));
> assert(capability_doesnt_contain_address(get_ddc(), x));
> set_ddc_to(<x..x+y>);
> assert(capability_doesnt_contain_address(get_ddc(), a));
> assert(capability_does_contain_address(get_ddc(), x));
> ```
> 
> Do those sound doable?

I don't like to just print stuff, either. Asserting seems doable. Regarding better tests for "SIGPROT", I remember there was no better choice when we decided for the `bash` scripts. I have to think about it.

hybrid/ddc_bounds.c
jacobbramley: I don't think you want `cheri_cap_build` here.

----

Loosely: `cheri_cap_build(x, y)` takes one bit pattern, `y`, that looks like a capability (but perhaps isn't tagged), and a valid capability `x`, and derives a new capability from `x` that looks like `y`, but valid and unsealed. It doesn't give you any capability you couldn't have already derived (since you have to have `x` in the first place).

This allows you to do two things, broadly:
* If you have a capablity `x`, and a sealed capability `y` that is otherwise a subset of `x`, you can unseal `y` (using `x` as credentials instead of an `UNSEAL` permission).
* If you have a capability `x` and an _invalid_ capability `y` that would have been a subset of `x` if it were valid, you can derive a capability from `x` that looks like `y` with its tag set.

Both operations could be achieved by a series of individual operations on `x` (e.g. `cheri_bounds_set()` and so on).

drpdr: > It depends on what you want to do. What bounds do you expect `array_cap` to have? `array_ptr` is a just a plain address, and the C type has no size information. I doubt that the compiler will attempt to preserve the bounds of `array` across those casts (though in purecap it would because `&array` would itself be a capability).
> 
> If your intention is to restrict the bounds of `array_cap` to the array itself, then you probably need to use `cheri_bounds_set` to do that explicitly. (Check that, though, because I haven't.)

If I try to set the bounds on the capability, when I do a `write_ddc` I get a `Capability bounds fault`. So, there is something I haven't got right here...

hybrid/include/utils.h
```diff
@@ -0,0 +1,25 @@
...
+void *__capability read_ddc();
+
+/* Write a capability to the DDC
+ */
+inline void write_ddc(void *__capability cap)
+{
+	asm("MSR DDC, %[cap]\n\t"
+		:
+		: [cap] "C"(cap) /* input regs */
+		:);
```
jacobbramley, line 15: I think you want a "memory" clobber here, since you're writing a system register that the compiler doesn't know about.

As discussed, this gets complicated in general (e.g. Laurie's benchmarking example) but it works in my local test framework.

jacobbramley: Alternatively (and I strongly recommend this here), make this whole file `utils.s` assembly file and just rule out compiler interference that way.

drpdr: I tried "memory" yesterday but it didn't work. Your `utils.s` solution maybe it is a better way to address this issue. Assuming, there is no other interference or problem `CHERI/builtin`-related.

drpdr: I was wrong, this now works in 1ae5ff2.

hybrid/ddc_invalid.c
```diff
@@ -0,0 +1,19 @@
+/***
+ * This program causes an "In-address space security Exception"
+ * by clearing the tag of the capability in the DDC
```
ltratt, line 3: Where is the exception caused? It might be necessary to split line 17 into separate parts so that you can be clear about which causes the exception.

hybrid/ddc_null.c
```diff
@@ -0,0 +1,20 @@
+/***
+ * This programs shows that putting
+ * "NULL" into the DDC causes an
```
ltratt, line 3: Same comment as above: where is the exception caused?

hybrid/hello_ddc.c
```diff
@@ -0,0 +1,25 @@
...
+
+int main()
+{
+	pp_cap(read_ddc());
+	uint16_t *some_int_ptr = (uint16_t *) malloc(sizeof(uint16_t));
+	*some_int_ptr = 200;
+	write_ddc((void *__capability) some_int_ptr);
+	uint16_t *__capability ddc_val = (uint16_t *__capability) cheri_ddc_get();
+	pp_cap(cheri_ddc_get());
+	printf("Value in DDC: %d\n", *ddc_val);
```
ltratt, line 23: It would be good to make this an `assert`.

hybrid/include/utils.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+void write_ddc(void *__capability cap);
+void *__capability read_ddc();
+
+/* Write a capability to the DDC
+ */
+inline void write_ddc(void *__capability cap)
+{
+	asm("MSR DDC, %[cap]\n\t"
```
ltratt, line 12: Should this function/file be guarded as only compileable on Morello?

hybrid/hello_ddc.c
drpdr: Yes, as you have suggested we will use assertions instead of prints in every file.

hybrid/include/utils.h
drpdr: Yes, better here than in every file.

drpdr:
> @drpdr Can we create (or maybe better, adjust the existing?) "tests" that `assert` that things will be one way or the other rather than just printing values out? For example (in pseudocode) I'd like to see a test something like:
> 
> ```
> x = malloc(y);
> set_ddc_to(<something that doesn't cover x..x+y>);
> assert(capability_doesnt_contain_address(get_ddc(), x));
> ```
> 
> The advantage of `assert` tests is that they integrate well with CI. The disadvantage is that they don't test that a program actually crashes when it does something naughty, so we also need some "this actually SIGPROTs" tests too.
> 
> Another test I'd like to see is something like:
> 
> ```
> x = malloc(y);
> assert(capability_does_contain_address(get_ddc(), x));
> a = malloc(b);
> set_ddc_to(<a..a+b>);
> assert(capability_does_contain_address(get_ddc(), a));
> assert(capability_doesnt_contain_address(get_ddc(), x));
> set_ddc_to(<x..x+y>);
> assert(capability_doesnt_contain_address(get_ddc(), a));
> assert(capability_does_contain_address(get_ddc(), x));
> ```
> 
> Do those sound doable?

Addressed in 754835, I *think*. We'll need some refactoring after this PR to make assertions consistent with the rest of the examples. Ignore `ddc_bounds`. It has not been covered yet for the issue above I am still investigating.

drpdr:
Added more comments in `hello_ddc`, in 886308f.

hybrid/ddc_invalid.c
drpdr: This should have been fixed by the assertions in 7548354.

hybrid/ddc_null.c
drpdr: Assertions in 7548354.

hybrid/ddc_invalid.c
```diff
@@ -14,6 +14,9 @@
 int main()
 {
 	pp_cap(cheri_ddc_get());
+	// Before clearing the tag we ensure we have a valid one
+	assert(cheri_address_get(cheri_ddc_get()) == 0 && cheri_tag_get(cheri_ddc_get() == 1));
```
ltratt, line 18: `cheri_tag_get(cheri_ddc_get() == 1)` doesn't look right -- should it be `cheri_tag_get(cheri_ddc_get()) == 1` or, perhaps more idiomatically, just `cheri_tag_get(cheri_ddc_get())`?

hybrid/ddc_null.c
```diff
@@ -15,6 +16,9 @@
 int main()
 {
 	pp_cap(cheri_ddc_get());
+	// The DDC is empty and we ensure it is
```
ltratt, line 19: What does "The DDC is empty" mean? I assume it means "the DDC is NULL"? And I guess that can only happen if this file is compiled in purecap mode?

hybrid/hello_ddc.c
```diff
@@ -14,12 +14,14 @@
...
-	write_ddc((void *__capability) some_int_ptr);
 	uint16_t *__capability ddc_val = (uint16_t *__capability) cheri_ddc_get();
-	pp_cap(cheri_ddc_get());
-	printf("Value in DDC: %d\n", *ddc_val);
+	pp_cap(ddc_val);
+	assert(cheri_address_get(ddc_val) != some_int_ptr);
+	write_ddc((void *__capability) some_int_ptr);
+	ddc_val = (uint16_t *__capability) cheri_ddc_get();
+	assert(cheri_address_get(ddc_val) == some_int_ptr);
+	pp_cap(ddc_val);
```
ltratt, line 27: Do we need the `pp_cap`s? I think we can kill them?

ltratt:
> It has not been covered yet for the issue above I am still investigating.

OK.

hybrid/ddc_null.c
drpdr: Well, if NULL it would crash. In hybrid-mode when we haven't put anything inside it, it's address is all zeros. It contains a valid tag and address with all zeros.

hybrid/hello_ddc.c
drpdr: Sure, we'll do.

hybrid/ddc_null.c
ltratt: "All zeros" is NULL on the platforms we care about.

> And I guess that can only happen if this file is compiled in purecap mode?

Can I ping you on this bit too please.

drpdr: Here we are in hybrid mode, so we are not talking about purecap. If what you said were true, the assertions would fail. This example is hybrid mode only.

ltratt: OK let's either a) add a compile time `#ifdef`-esque check that we're not in purecap mode (preferably) or b) at least document that this is hybrid only. Can we do that for all the examples please?

drpdr: All the examples here were developed for hybrid mode. If you mean the whole repo, of course, we can do it. Here were are experimenting in hybrid mode, unless I have missed some pieces along the way...

drpdr: And I'll document this better.

ltratt: For the files in this PR, let's add compile-time checks (if possible -- which I think it is), as they're the best form of documentation! It doesn't hurt to add human-friendly documentation to the same effect too, of course!

drpdr: > "All zeros" is NULL on the platforms we care about.
> 
> > And I guess that can only happen if this file is compiled in purecap mode?
> 
> Can I ping you on this bit too please.

This is how the DDC looks when compiling in hybrid mode and printing is content at the beginning of the program:
```
Capability: 0x0 [rwRW,0x0-0x1000000000000]
Tag: 1, Perms: 3717d, Type: 0, Address: 0000, Base: 0000, End: 1000000000000, Flags: 0, Length: 1000000000000, Offset: 0000
```

As you can see, the capability is valid but it's address is zero. When we put a `NULL` into it, the capability will become *invalid*. You wanted to clarify this point further, so I thought this would have been a good place.

ltratt: This isn't a NULL capability: it has a NULL address but notice that the `End` is != 0. In other words, it's more-or-less the DDC I would have expected: its bounds cover what I assume is the entirety of the virtual address?

drpdr: > This isn't a NULL capability: it has a NULL address but notice that the `End` is != 0. In other words, it's more-or-less the DDC I would have expected: its bounds cover what I assume is the entirety of the virtual address?

I had referred to this as "EMPTY" but, of course, is a wrong naming. I *think* you are right regarding its bounds.

ltratt: Anyway, let's please put the `#ifdef`-esque checks that mean that we're only compiling this in hybrid mode. It should be easy to do (I think!) and will help remind our future selves what's going on in the files that result from this PR.

drpdr: > Anyway, let's please put the `#ifdef`-esque checks that mean that we're only compiling this in hybrid mode. It should be easy to do (I think!) and will help remind our future selves what's going on in the files that result from this PR.

OK.

drpdr: > Anyway, let's please put the `#ifdef`-esque checks that mean that we're only compiling this in hybrid mode. It should be easy to do (I think!) and will help remind our future selves what's going on in the files that result from this PR.

Are you thinking about custom macro definition to check a capability or something different? Because I can't find anything in the `cheriintrin.h`...

ltratt: Something like https://github.com/capablevms/cheri_misidioms/blob/master/code/narrow.c#L7 (but not purecap!).

drpdr: > Something like https://github.com/capablevms/cheri_misidioms/blob/master/code/narrow.c#L7 (but not purecap!).

It seems the only way (they also use only the purecap check in all the projects). There is no equivalent like "*_HYBRID". So checking we are in purecap but we should not be (since this PR is hybrid only), I think should be sufficient.

drpdr: Fix in 28ade2d.

hybrid/include/utils.h
drpdr: Fix in fea1493.

hybrid/hello_ddc.c
drpdr: Fix in bcc4c6d.

hybrid/ddc_null.c
ltratt: Technically this will compile on any normal architecture. I *think* we can get "CHERI hybrid mode" with something like:

```c
#if !defined(__CHERI_CAPABILITY_WIDTH__)
#error "Only works on CHERI"
#elif defined(__CHERI_PURE_CAPABILITY__)
#error "Only works on CHERI hybrid mode"
#endif
```

We could squish that down to:

```c
#if defined(__CHERI_CAPABILITY_WIDTH__) || !defined(__CHERI_PURE_CAPABILITY__)
#error "Only works on CHERI hybrid mode"
#endif
```

I think.

drpdr: You are quite right.

drpdr: > Technically this will compile on any normal architecture. I _think_ we can get "CHERI hybrid mode" with something like:
> 
> ```c
> #if !defined(__CHERI_CAPABILITY_WIDTH__)
> #error "Only works on CHERI"
> #elif defined(__CHERI_PURE_CAPABILITY__)
> #error "Only works on CHERI hybrid mode"
> #endif
> ```
> 
> We could squish that down to:
> 
> ```c
> #if defined(__CHERI_CAPABILITY_WIDTH__) || !defined(__CHERI_PURE_CAPABILITY__)
> #error "Only works on CHERI hybrid mode"
> #endif
> ```
> 
> I think.

It seems `__CHERI_CAPABILITY_WIDTH__` (128) is defined even on *hybrid mode*.

ltratt: > It seems __CHERI_CAPABILITY_WIDTH__ (128) is defined even on hybrid mode.

Oops, I've got my logic the wrong way around! It should be:

```c
#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
#error "Only works on CHERI hybrid mode"
#endif
```

hybrid/ddc_invalid.c
drpdr: I can't reply to your [comment](https://github.com/capablevms/cheri-examples/pull/39#commitcomment-57175418) below but it is also related to checking the tag. Fix in 9a33253.

hybrid/hello_ddc.c
drpdr: Addressed in 108c268.

hybrid/ddc_null.c
drpdr: Addressed in 108c26a.

ltratt:
This one's gotten lost in the noise:


>Another test I'd like to see is something like:
>
>x = malloc(y);
assert(capability_does_contain_address(get_ddc(), x));
a = malloc(b);
set_ddc_to(<a..a+b>);
assert(capability_does_contain_address(get_ddc(), a));
assert(capability_doesnt_contain_address(get_ddc(), x));
set_ddc_to(<x..x+y>);
assert(capability_doesnt_contain_address(get_ddc(), a));
assert(capability_does_contain_address(get_ddc(), x));

I'd still like to see it in this PR :)

drpdr:
> This one's gotten lost in the noise:
> 
> > Another test I'd like to see is something like:
> > x = malloc(y);
> > assert(capability_does_contain_address(get_ddc(), x));
> > a = malloc(b);
> > set_ddc_to(<a..a+b>);
> > assert(capability_does_contain_address(get_ddc(), a));
> > assert(capability_doesnt_contain_address(get_ddc(), x));
> > set_ddc_to(<x..x+y>);
> > assert(capability_doesnt_contain_address(get_ddc(), a));
> > assert(capability_does_contain_address(get_ddc(), x));
> 
> I'd still like to see it in this PR :)

Probably the best place to put this is in `hello_ddc.c`. What do you think?

ltratt:
> Probably the best place to put this is in hello_ddc.c. What do you think?

I *think* maybe we should remove the current `hello_ddc.c` and make a new file `basic_ddc.c` (or some similar name that doesn't suggest it'll print "hello world") with the `assert` sequence I suggested. I don't feel strongly wedded to this, however.

drpdr:
> > Probably the best place to put this is in hello_ddc.c. What do you think?
> 
> I _think_ maybe we should remove the current `hello_ddc.c` and make a new file `basic_ddc.c` (or some similar name that doesn't suggest it'll print "hello world") with the `assert` sequence I suggested. I don't feel strongly wedded to this, however.

I see. Well my `hello` was more like *hello I am DDC and I can do this*, but `basic_ddc.c` sounds appropriate.

drpdr:
> > > set_ddc_to(<a..a+b>);

I am not sure I have got this right. `a` and `a+b` are... addresses? Do you want `a` to be extended to `a+b` first and then set the DDC with that capability?

ltratt:
Vaguely: `a` == "base" and `b` == "length".

drpdr:
> > Probably the best place to put this is in hello_ddc.c. What do you think?
> 
> I _think_ maybe we should remove the current `hello_ddc.c` and make a new file `basic_ddc.c` (or some similar name that doesn't suggest it'll print "hello world") with the `assert` sequence I suggested. I don't feel strongly wedded to this, however.

This should have been fixed in 3b5eaaa.

drpdr:
> > > Probably the best place to put this is in hello_ddc.c. What do you think?
> > 
> > 
> > I _think_ maybe we should remove the current `hello_ddc.c` and make a new file `basic_ddc.c` (or some similar name that doesn't suggest it'll print "hello world") with the `assert` sequence I suggested. I don't feel strongly wedded to this, however.
> 
> I see. Well, my `hello` was more like _hello I am DDC and I can do this_, but `basic_ddc.c` sounds appropriate.

Renamed in c62b0fc.

drpdr:
In 3b29901, I've tried to restrict the bounds from the initial valid capability of the DDC. As we discussed in the meeting @jacobbramley.

hybrid/hello_ddc.c
```diff
@@ -1,6 +1,6 @@
 /***
- * This program creates a capability and writes
- * it to the Default Data Capability register.
+ * This program does some basic reads/writes
```
ltratt: Let's expand this description: what this program really does is malloc two blocks and flip the DDC to point from one to the other. In other words, it shows (at a very very very basic level) compartmentalisation.

drpdr: Done in 0f82fad.

ltratt:
My suggestion for this PR is that we remove `ddc_bounds.c` from the PR entirely, as we're going to have to do something more complex for that case, and I think it would make more sense to do that in a new PR.

drpdr:
I think this PR can now leave draft state and it is so ready for review. I have also updated its description.

hybrid/basic_ddc.c
```diff
@@ -0,0 +1,31 @@
+/***
+ * This programs shows a very basic level
+ * of comportmentalisation. It first `malloc`
```
ltratt, line 3: s/comportmentalisation/compartmentalisation/

hybrid/basic_ddc.c
```diff
@@ -0,0 +1,31 @@
...
+	*some_int_ptr = 200;
+	assert(cheri_address_get(cheri_ddc_get()) != some_int_ptr);
+	uint32_t *some_other_int_ptr = (uint32_t *) malloc(sizeof(uint32_t));
+	write_ddc((void *__capability) some_other_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) != some_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) == some_other_int_ptr);
+	write_ddc((void *__capability) some_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) != some_other_int_ptr);
+	assert(cheri_address_get(cheri_ddc_get()) == some_int_ptr);
+	return 0;
```
ltratt, line 30: I think it's the case that if `main` was another (non-inlined) function `f` then the program would crash at this `return`? If so, we should add to the description something like "Note: this program is very simple and writing to the DDC in this fashion would cause a crash if the program were to execute much further."

ltratt:
Can we also format all the comments to the same width (ideally 80 or 100 chars)? Personally I find it *much* easier to use `//` commenting exclusively for this (`/* ... */` feels very retrograde to me I must admit!).

hybrid/basic_ddc.c
drpdr: Oh, yes, as we know now.

hybrid/basic_ddc.c
drpdr: Fix in 239cfd0.

drpdr:
> Can we also format all the comments to the same width (ideally 80 or 100 chars)? Personally I find it _much_ easier to use `//` commenting exclusively for this (`/* ... */` feels very retrograde to me I must admit!).

So IIUC you would like to have all the `/* */` replaced with `//` but not the `/** **/`?

ltratt:
> So IIUC you would like to have all the /* */ replaced with //

Yes please.

hybrid/basic_ddc.c
drpdr: Added in 676d41a.

drpdr:
> > So IIUC you would like to have all the /* */ replaced with //
> 
> Yes please.

Done in c77119d.

drpdr:
With all the comments addressed this PR should be ready for re-review. I had forgotten to submit this comment.

ltratt:
Please squash.

drpdr:
> Please squash.

Squashed.

ltratt:
I think the first line of the commit might be messed up? It seems to be " Some simple interactions with the Default Data Capabiity register\non morello-hybrid-mode". The first line of a commit must be short unfortunately :/ Please force push an update (if I'm right!).

drpdr:
> I think the first line of the commit might be messed up? It seems to be " Some simple interactions with the Default Data Capabiity register\non morello-hybrid-mode". The first line of a commit must be short unfortunately :/ Please force push an update (if I'm right!).

You are right. Amended in 8047aca.

ltratt:
s/Capabiity/Capability/ in the commit title. Also, I think you should mention Morello in the commit body.

Please force push an update.

drpdr:
> s/Capabiity/Capability/ in the commit title. Also, I think you should mention Morello in the commit body.
> 
> Please force push an update.

Done.

PR #44, opened 2021-10-01
Draft PR about compartment switching

drpdr:
This is a simple interface so we can start discussing about what to do. Assembly trampoline not implemented yet.

### DIFF
--- a/hybrid/ddc_compartment_switching.c
+++ b/hybrid/ddc_compartment_switching.c
@@ -1,0 +1,36 @@
+/***
+ * TODO: something like intermediate level of compartment switching or ...
+ ***/
+
+#include "../include/common.h"
+#include "include/utils.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+// Simple function with some data which will be part of the compartment
+int compartment_simple_fun(int a, int b);
+// Temp hack to "design" an interface
+int switch_compartment(int a, int b, void * stack, size_t size);
+// TODO: this will call assembly code
+// extern int switch_compartment(int a, int b, void * stack, size_t size);
+
+int main(){
+    void * simple_block = malloc(5000);
+    size_t compartment_size = 2000;
+    switch_compartment(5, 3, simple_block, compartment_size);
+    return 0;
+}
+
+int switch_compartment(int a, int b, void * stack, size_t size){
+    return compartment_simple_fun(a, b); // FIXME: this is a temp hack
+}
+
+int compartment_simple_fun(int a, int b) {
+   return a + b;
+}


ltratt:
What does and doesn't work in this PR at the moment?

drpdr:
> What does and doesn't work in this PR at the moment?

This is a simple interface. Now, it does not flip the DDC. All the DDC part must be done via an assembly trampoline. There is  a little "hack" to compile and run the code. As you can see, the `extern` function is commented. That will be the one doing the job.

jacobbramley:
On my model:

* The [initial DDC](https://www.morello-project.org/capinfo/?c=0x1dc5f4000000300050000000000000000) is permissive.
* The [DDC in `compartment_simple_fun`](https://www.morello-project.org/capinfo/?c=0x1dc5f400057d090000000000040a09000) is restricted to the requested 2000 bytes.
* The original DDC is restored on exit.

hybrid/ddc_compartment_switching/shared.s
```diff
@@ -0,0 +1,141 @@
...
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+.global compartment_simple_fun
+
+.text
+.balign 4
+.global switch_compartment
+.type switch_compartment, "function"
+switch_compartment:
+    // int switch_compartment(int a, int b, void * stack, size_t size)
```
ltratt, line 11: Would I be silly if I said that we don't need `a, b` as arguments here (in the sense that they don't have anything to do with compartment switching as such)? Or is the intention here to model a slightly simplified mega-function `switch_compartment_and_call(new_stack, new_stack_size, function_f, function_f_arg_0, ... function_f_arg_n)`?

drpdr: I would have assumed `a` and `b` being the "content" but I am now trying to figure out how to show that "this works".

ltratt: I think at first it might be easiest to not have `a` and `b` in here at least at first (to keep it simple): when we switch compartments, the code can read whatever state it wants from the new heap.

drpdr:
Question: how do we know/show/prove that we are actually switching? What would we expect to happen when printing the content of the DDC in `compartment_simple_fun`?

ltratt:
We will need (at least) two separate programs: one which reads from the correct heap before/after switching; and another which reads from the incorrect heap before/after switching. The latter should SIGPROT (or similar) and our testing infrastructure will need to ensure that it fails.

drpdr:
> We will need (at least) two separate programs: one which reads from the correct heap before/after switching; and another which reads from the incorrect heap before/after switching. The latter should SIGPROT (or similar) and our testing infrastructure will need to ensure that it fails.

What do you mean by *correct* and *incorrect* heap? And as two separate programs, you mean two programs using the same compartment switching function (calling it API seems a bit too much here)?

drpdr:
> > We will need (at least) two separate programs: one which reads from the correct heap before/after switching; and another which reads from the incorrect heap before/after switching. The latter should SIGPROT (or similar) and our testing infrastructure will need to ensure that it fails.
> 
> What do you mean by _correct_ and _incorrect_ heap? And as two separate programs, you mean two programs using the same compartment switching function (calling it API seems a bit too much here)?

Oh, you mean the allocated 2000 VS 3000, the first 2000 used for the compartment switching...

hybrid/ddc_compartment_switching/shared.s
jacobbramley: Actually there is no "new heap", because there's no easy way to temporarily ask `malloc` to allocate from a different area. The switcher just sets up a stack, for now. We could implement a simple allocator, but that would dominate the example.

Perhaps a second example could incorporate @jeremysinger's bump pointer allocator or similar.

jacobbramley:
> Question: how do we know/show/prove that we are actually switching? What would we expect to happen when printing the content of the DDC in `compartment_simple_fun`?

Hence [this comment](https://github.com/capablevms/cheri-examples/pull/44#issuecomment-933299351). You could simple test that the DDC has the required length.

As a test, you could pass a pointer into the compartment switch (either to the caller's stack or its heap), and check that the callee can't access it.

ltratt:
I like `assert`y tests, and I agree we can use them here, but I think in this case we also need "crash"y tests to prove that the hardware really is borking when we think it should. I think this means that my "two (at least)" comment is safe -- we need at least three tests? :p

drpdr:
> I like `assert`y tests, and I agree we can use them here, but I think in this case we also need "crash"y tests to prove that the hardware really is borking when we think it should. I think this means that my "two (at least)" comment is safe -- we need at least three tests? :p

I have been trying to understand where we should `assert`. I can think of only one place: inside `compartment_simple_fun`. However, I can only cause crashes whatever I try to do after reading the DDC. I am sure I am missing something here (not CHERI-related).

jacobbramley:
> I have been trying to understand where we should `assert`. I can think of only one place: inside `compartment_simple_fun`. However, I can only cause crashes whatever I try to do after reading the DDC. I am sure I am missing something here (not CHERI-related).

* Does `assert` allocate and use heap memory? If so, it will crash.
    * It might use this for printing, for example.
* Does `assert` trigger any unwinding logic? This will fail too.

hybrid/ddc_compartment_switching/shared.s
jacobbramley: By the way, implementing a prototype-agnostic `switch_compartment_and_call` is much more difficult because it needs to do general-case argument marshalling.

drpdr:
> Hence [this comment](https://github.com/capablevms/cheri-examples/pull/44#issuecomment-933299351). You could simply test that the DDC has the required length.

But, "The DDC in compartment_simple_fun is restricted to the requested 2000 bytes.". So, a `cheri_ddc_get()` will give the  capability but crash when we try to access it to read its length, etc. ?

drpdr:
>   * It might use this for printing, for example.
I had't thought about allocation and printing. A simple `printf` in `compartment_simple_fun` will crash. Right.

drpdr:
> I like `assert`y tests, and I agree we can use them here, but I think in this case we also need "crash"y tests to prove that the hardware really is borking when we think it should. I think this means that my "two (at least)" comment is safe -- we need at least three tests? :p

1) `assert` we have the right sized compartment, e.g. 2000 bytes;
2) trying to access the heap will cause a crash in `compartment_simple_fun`, e.g. a simple `printf`;
3) ?

ltratt:
I'd prefer doing just a simple load or store rather than a full-on `printf` (which might work for other reasons not related to our compartmentalisation). I think our 2 (or more...) tests should be:

1. Create two compartments *A* and *B*, switch between them, loading and storing in *A* and *B* after switch.
2. Create two compartments *A* and *B*, switch between them, load or store from the other compartment (and check it SIGPROTs).

drpdr:
> I'd prefer doing just a simple load or store rather than a full-on `printf` (which might work for other reasons not related to our compartmentalisation). I think our 2 (or more...) tests should be:
> 
> 1. Create two compartments _A_ and _B_, switch between them, loading and storing in _A_ and _B_ after switch.
> 2. Create two compartments _A_ and _B_, switch between them, load or store from the other compartment (and check it SIGPROTs).

And just to double check as compartment in our case we intend the first 2000 plus 3000 of our total of 5000 bytes. The function can access only the first 2000 and it will crash otherwise.

drpdr:
How can I get a valid capability `cap` in hybrid context to pass to `switch_compartment(void *__capability cap, void * stack, size_t size)`?

ltratt:
> How can I get a valid capability

That's not a concern at the moment: let's just assume a compartment might have access to such a capability (however it got it).

drpdr:
> > How can I get a valid capability
> 
> That's not a concern at the moment: let's just assume a compartment might have access to such a capability (however it got it).

But if I cannot pass a valid capability I cannot run a non-aborting test. If the capability is not valid, program will crash on calling the function.

ltratt:
Let's take a step back for this first PR: let's get rid of the arguments `a, b` and not replace them with anything (so we can't pass any values to, or receive values from, a compartment switch). That should make it clearer what's going on at a basic level.

I will raise an issue where I sketch out what I think is a more promising medium term API and we can tackle that in the next PR.

drpdr:
I've tried many other things but let's try to figure out what is going on with very small steps. In ee2bf1d, I wanted just to remove arguments `a` and `b` (before trying to use `c0`). However, I have a `bus error` when copying the `sp`. Is `w0` still the return value?

ltratt:
Can you give line number(s) for the errors please/

drpdr:
> Can you give line number(s) for the errors please?

https://github.com/capablevms/cheri-examples/blob/ee2bf1d75862add6c4d78fe5a030c88e4479f60b/hybrid/ddc_compartment_switching/shared.s#L46

ltratt:
You've changed lines after line 46, so I think you might be looking in the wrong place?

drpdr:
> You've changed lines after line 46, so I think you might be looking in the wrong place?

I understand. But I have given you the line number for the error (SIGBUS). In other words, removing the arguments and register `mov w0 ... w1` makes the program crash on that line.

ltratt:
Something doesn't make sense: why would changing line 70-odd cause a crash at line 46? Is it happening the second time you execute line 46 or ... ?

drpdr:
> Something doesn't make sense: why would changing line 70-odd cause a crash at line 46? Is it happening the second time you execute line 46 or ... ?

If I put the arguments back, everything works. You are right. It doesn't make sense. The error must be elsewhere.

ltratt:
Right, so I think you need to start tracing execution. Maybe set a breakpoint at line 46-ish to see if it's being executed more than once etc.

jacobbramley:
At a glance: Removing the first two arguments (`w0` and `w1`) means that the second two arguments (`stack` and `size`) now end up in `x0` and `x1`, respectively, but I don't think the patch includes changes to account for that.

This is my fault; a better design would put them first in the list so that we can experiment with the real arguments without these changes.

drpdr:
> At a glance: Removing the first two arguments (`w0` and `w1`) means that the second two arguments (`stack` and `size`) now end up in `x0` and `x1`, respectively, but I don't think the patch includes changes to account for that.
> 
> This is my fault; a better design would put them first in the list so that we can experiment with the real arguments without these changes.

Oh, no, I should have thought about the arguments better. You cannot simply remove stuff and expect the following args to stay where they were before. Thanks for the explanation.

drpdr:
> I've tried many other things but let's try to figure out what is going on with very small steps. In [ee2bf1d](https://github.com/capablevms/cheri-examples/commit/ee2bf1d75862add6c4d78fe5a030c88e4479f60b), I wanted just to remove arguments `a` and `b` (before trying to use `c0`). However, I have a `bus error` when copying the `sp`. Is `w0` still the return value?

This is now fixed in 17b23f5. I can get back to the PR and tests.

ltratt:
Excellent! BTW, should we add something to `.buildbot.sh` to make sure we at least try building the new file(s)?

drpdr:
Let me try to clarify my thoughts. Now we have just an assertion after this [ line](https://github.com/capablevms/cheri-examples/blob/17b23f5fd50ff474a03f8eda5016e52e05e50457/hybrid/ddc_compartment_switching/ddc_compartment_switching.c#L34) which we know it works because we know how big is the compartment (2000). If we state something different (not exact) like `length == 2001 or 1999` it aborts. 
On that line I obtain a capability from the DDC, which is authentic so I should be able to dereference it and assert what it points to, right? E.g.`ddc_cap[1000] == 100`, assuming I have put `100` in that block. I must be very precise, of course, otherwise, as I said above, it will crash. I do not need to get its address because even in hybrid mode I should be able to just dereference an authentic capability

ltratt:
> On that line I obtain a capability from the DDC, which is authentic so I should be able to dereference it and assert what it points to, right?

Not necessarily: the DDC can be authentic but still point to unmap'd memory. At a high-level I could do:

```c
(uint8_t *) x = malloc(4096);
set_ddc_to(x, 4096);
free(x);
// the DDC is still authentic at this point but the next line should crash
*x = 1;
```

drpdr:
> Excellent! BTW, should we add something to `.buildbot.sh` to make sure we at least try building the new file(s)?

Of course, will do.

drpdr:
> > On that line I obtain a capability from the DDC, which is authentic so I should be able to dereference it and assert what it points to, right?
> 
> Not necessarily: the DDC can be authentic but still point to unmap'd memory. At a high-level I could do:
> 
> ```c
> (uint8_t *) x = malloc(4096);
> set_ddc_to(x, 4096);
> free(x);
> // the DDC is still authentic at this point but the next line should crash
> *x = 1;
> ```

In general you are right but in our case the assertion would fail in case of unmap'd memory (?).

ltratt:
Which assertion are you referring to? I know I sound like a stuck record, but I need line numbers each time you mention a piece of code -- I have the memory of a peanut :)

drpdr:
> Which assertion are you referring to? I know I sound like a stuck record, but I need line numbers each time you mention a piece of code -- I have the memory of a peanut :)

ATM [this](https://github.com/capablevms/cheri-examples/blob/17b23f5fd50ff474a03f8eda5016e52e05e50457/hybrid/ddc_compartment_switching/ddc_compartment_switching.c#L35) works because we know that function can access the first 2000 bytes. Sorry for adding these details but we have this context here. So, that memory is mapped.

ltratt:
> assert(cheri_length_get(ddc_cap) == 2000)

It's worth remembering that an authentic capability is authentic *no matter what it's pointing at*. So you can have an authentic capability with an address and a length that points to invalid memory. [FWIW, the assert you pointed at doesn't guarantee authenticity. You probably want something like `assert(cheri_tag_get(ddc_cap) && cheri_length_get(ddc_cap) == 2000)` to double check that.]

Can you push a commit / add a comment which shows specifically a) the line(s) of code you're trying that unexpectedly causes a crash b) where those line(s) of code live in the overall example.

drpdr:
In f5bcb01, I am just tying to verify what is in compartment A and B so we then can develop separate tests and refactoring/clean up to make things clear.

ltratt:
Your problem is probably:

```c
	int *simple_block = malloc(5000);
	simple_block[1000] = 100;
```

Notice that `simple_block[1000]` means "the 1000th integer" which means you're doing `*((char *) simple_block + 1000 * 8) = 100` so you're reading past the end of the malloc'd block. You're probably at risk of both SIGPROT (you're *probably* reading past the capability's bound) and SIGSEGV (because you're *probably* past the end of the mapped physical page).

drpdr:
> > assert(cheri_length_get(ddc_cap) == 2000)
> 
> It's worth remembering that an authentic capability is authentic _no matter what it's pointing at_. So you can have an authentic capability with an address and a length that points to invalid memory. [FWIW, the assert you pointed at doesn't guarantee authenticity. You probably want something like `assert(cheri_tag_get(ddc_cap) && cheri_length_get(ddc_cap) == 2000)` to double check that.]
> 
> Can you push a commit / add a comment which shows specifically a) the line(s) of code you're trying that unexpectedly causes a crash b) where those line(s) of code live in the overall example.

This assertion causes an bounds vio

> Your problem is probably:
> 
> ```c
> 	int *simple_block = malloc(5000);
> 	simple_block[1000] = 100;
> ```
> 
> Notice that `simple_block[1000]` means "the 1000th integer" which means you're doing `*((char *) simple_block + 1000 * 8) = 100` so you're reading past the end of the malloc'd block. You're probably at risk of both SIGPROT (you're _probably_ reading past the capability's bound) and SIGSEGV (because you're _probably_ past the end of the mapped physical page).

I've just noticed this while writing another comment... Thanks!

drpdr:
> > Your problem is probably:
> > ```c
> > 	int *simple_block = malloc(5000);
> > 	simple_block[1000] = 100;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Notice that `simple_block[1000]` means "the 1000th integer" which means you're doing `*((char *) simple_block + 1000 * 8) = 100` so you're reading past the end of the malloc'd block. You're probably at risk of both SIGPROT (you're _probably_ reading past the capability's bound) and SIGSEGV (because you're _probably_ past the end of the mapped physical page).
> 
> I've just noticed this while writing another comment... Thanks!

Fixed in 5aabb8b. Now everything should be alright to make separate tests and verify what we said some comments above.

drpdr:
Is the naming here, i.e. `switching`, appropriate? Are we really `switching` or creating a small compartment, instead?

drpdr:
Two very similar programs (which share the trampoline) testing the presence of a simple `int` in compartment A or B. In 408f519. One passes the assertion (`ddc_compartment_switching.c`), the other (`ddc_*nok`) causes a `CHERI violation`.

hybrid/ddc_compartment_switching/ddc_compartment_switching_nok.c
```diff
@@ -0,0 +1,41 @@
...
+	return 0;
+}
+
+int compartment_simple_fun()
+{
+	int32_t *__capability ddc_cap = cheri_ddc_get();
+	// This function can access only 2000 bytes, i.e. `compartment_size`
+	assert(cheri_tag_get(ddc_cap) && cheri_length_get(ddc_cap) == 2000);
+	// This function cannot go over its compartment bounds
+	assert(ddc_cap[700]);
```
ltratt, line 39: This `assert` doesn't look right: either the derefence fails because it's out of bounds, or we're testing that `assert(8 != 0)` which the compiler might even be able to statically prove is true.

drpdr: Maybe a comment and single assertion could be better/more clear. In cb8307f.

drpdr:
> > Excellent! BTW, should we add something to `.buildbot.sh` to make sure we at least try building the new file(s)?
> 
> Of course, will do.

The failing test here will be similar to our other "In-address security exception" tests which we do not have anymore (???). I've just realised they have disappeared. I remember a discussion you had with Jacob because they did not add anything when executed on `qemu`. However, they add the `...OK/...NOK` output...

ltratt:
I don't remember wanting the failing tests to disappear, but I remind you that I have a memory of a peanut! We certainly want failing tests here (assuming it's possible).

drpdr:
> I don't remember wanting the failing tests to disappear, but I remind you that I have a memory of a peanut! We certainly want failing tests here (assuming it's possible).

I have to check better, but I think we do not have any test verifying the failure. So, I have to add them back, I guess...

ltratt:
We don't need to fix things not related to this PR in this PR (we can raise another PR to reenable those other failing tests). One thing at a time etc etc :)

drpdr:
> We don't need to fix things not related to this PR in this PR (we can raise another PR to reenable those other failing tests). One thing at a time etc etc :)

There is a `TODO` [here](https://github.com/capablevms/cheri-examples/blob/72584ff686d56095fad13731c00498d43791052f/tests/run_tests.sh#L20) I had missed.

jacobbramley:
> > > Excellent! BTW, should we add something to `.buildbot.sh` to make sure we at least try building the new file(s)?
> > 
> > 
> > Of course, will do.
> 
> The failing test here will be similar to our other "In-address security exception" tests which we do not have anymore (???). I've just realised they have disappeared. I remember a discussion you had with @jacobbramley because they did not add anything when executed on `qemu`. However, [they](https://github.com/capablevms/cheri-examples/blob/622f7e403da119c9118a47528af61493df7bd8fa/tests/run_tests.sh#L42) had the `...OK/...FAILED` output...

Yep, when I replaced the makefiles I found that whilst the tests were run, their outcome was ignored, and since some tests expect to fail, fixing that was non-trivial and seemed out of scope of that change.

drpdr:
> > > > Excellent! BTW, should we add something to `.buildbot.sh` to make sure we at least try building the new file(s)?
> > > 
> > > 
> > > Of course, will do.
> > 
> > 
> > The failing test here will be similar to our other "In-address security exception" tests which we do not have anymore (???). I've just realised they have disappeared. I remember a discussion you had with @jacobbramley because they did not add anything when executed on `qemu`. However, [they](https://github.com/capablevms/cheri-examples/blob/622f7e403da119c9118a47528af61493df7bd8fa/tests/run_tests.sh#L42) had the `...OK/...FAILED` output...
> 
> Yep, when I replaced the makefiles I found that whilst the tests were run, their outcome was ignored, and since some tests expect to fail, fixing that was non-trivial and seemed out of scope of that change.

Yes, when we thought about those tests we decided was enough to check the output, IIRC. The python `cheribuild qemu test-runner` does not check our outputs, of course. I remember the discussion but then I forgot about that...

drpdr:
> Excellent! BTW, should we add something to `.buildbot.sh` to make sure we at least try building the new file(s)?

In 94c8d2d.

drpdr:
I renamed and rephrased the title and description. Probably, we also need to rename the two examples with more appropriate terms (?). I think now this PR can leave draft state.

hybrid/ddc_compartment_switching/ddc_compartment_switching_nok.c
```diff
@@ -34,8 +34,7 @@ int compartment_simple_fun()
 {
 	int32_t *__capability ddc_cap = cheri_ddc_get();
 	// This function can access only 2000 bytes, i.e. `compartment_size`
-	assert(cheri_tag_get(ddc_cap) && cheri_length_get(ddc_cap) == 2000);
-	// This function cannot go over its compartment bounds
-	assert(ddc_cap[700]);
+	// So the following will go over its compartment bounds
+	assert(ddc_cap[700] == 8);
```
ltratt, line 38: What happens when you run this line? Does it SIGPROT?

drpdr: Yes.

ltratt: Then the `assert` is misleading: it's never executed, right?

drpdr: If we remove the `assert` it does not SIGPROT, though.

jacobbramley: > Then the `assert` is misleading: it's never executed, right?

Why would the `assert` not be executed? The access is out of bounds, and both versions try to load from it. (The old assertion is just `ddc_cap[700] != 0`.)

ltratt:
OK let's wait for comments from @jacobbramley.

drpdr:
I am not sure @jacobbramley got the notification even though I know you are very busy, Jacob...

hybrid/ddc_compartment_switching/ddc_compartment_switching.c
```diff
@@ -0,0 +1,42 @@
+/***
+ * This example shows an intermediate level of compartment creation.
+ * It allocates a block of memory (5000 bytes) but only the first
+ * 2000 are accessible by `compartment_simple_fun`.
+ *
+ * This program does not rise any exception because we verify we can
```
jacobbramley, line 6: s/rise/raise/

hybrid/ddc_compartment_switching/shared.s
```diff
@@ -0,0 +1,138 @@
...
```
jacobbramley, line 11: It'd be useful to update this comment, since it no longer reflects the C prototype.

hybrid/ddc_compartment_switching/shared.s
```diff
@@ -0,0 +1,138 @@
...
+    scvalue   c11, c10, x0
+    scbndse   c11, c11, x1
+    msr       DDC, c11
+
+    // Replace the stack pointer.
+    mov       x12, sp
+    add       sp, x0, x1
+
+    // Save the old DDC, stack pointer and return address on the new stack, so
+    // we can restore it when we return..
```
jacobbramley, line 50: There are two full stops. (I think that's my fault.)

hybrid/ddc_compartment_switching/shared.s
```diff
@@ -0,0 +1,138 @@
...
+    //            sp +  0 -> [ old DDC (low 64)  ]   |
+    //                                 :             :
+    //            `stack` -> ________________________v
+    //
+    // Note that this is _not_ an AAPCS64 frame record, even though it looks a
+    // bit like one. We don't touch FP here, and since it is not a capability
+    // (in hybrid mode), unwinding would fail anyway.
+
+    // Clean capabilities left in arguments.
+
```
jacobbramley, line 72: This comment should go, now that we no longer have arguments to clean.

hybrid/ddc_compartment_switching/shared.s
```diff
@@ -0,0 +1,138 @@
...
+    //                                 :             :
+    //            `stack` -> ________________________v
+    //
+    // Note that this is _not_ an AAPCS64 frame record, even though it looks a
+    // bit like one. We don't touch FP here, and since it is not a capability
+    // (in hybrid mode), unwinding would fail anyway.
+
+    // Clean capabilities left in arguments.
+    
+    bl (clean + 8)
```
jacobbramley, line 73: Just `bl clean`, since we no longer need to preserve arguments.

jacobbramley (commented):
Minor comments only. Otherwise, it looks fine.

hybrid/ddc_compartment_switching/shared.s
drpdr: Fix in 6284b24.

jacobbramley (commented):
Looks good to me (but @ltratt should approve since I wrote some of the code).

drpdr:
> Looks good to me (but @ltratt should approve since I wrote some of the code).

Yes, I'll be sure to keep your commit at squash time. I think I have to keep the first, the second, add a third and *then* squash. Waiting for @ltratt 's approval, of course.

hybrid/ddc_compartment_switching/ddc_compartment_switching.c
```diff
@@ -0,0 +1,42 @@
+/***
+ * This example shows an intermediate level of compartment creation.
+ * It allocates a block of memory (5000 bytes) but only the first
+ * 2000 are accessible by `compartment_simple_fun`.
+ *
+ * This program does not raise any exception because we verify we can
```
ltratt, line 6: This sentence isn't quite right (it's not "because we verify the load"). "This program does not raise an exception because it explicitly performs loads/stores within the compartment." might be better?

hybrid/ddc_compartment_switching/ddc_compartment_switching_nok.c
```diff
@@ -0,0 +1,40 @@
+/***
+ * This example shows an intermediate level of compartment creation.
+ * It allocates a block of memory (5000 bytes) but only the first
+ * 2000 are accessible by `compartment_simple_fun`.
+ *
+ * This program raises an exception because we try to access an
+ * `int32_t` which occurs to be out of the compartment's bounds.
```
ltratt, line 7: s/which occurs to be/which is/

ltratt:
In retrospect, I think it would be better to `malloc memory for `uint8_t`s and not `u32`s -- it makes it easier to understand which are correct accesses or not (without having to multiple integer sizes in one's head). Can we make that change please @drpdr?

hybrid/ddc_compartment_switching/ddc_compartment_switching.c
drpdr: Done in c7507be.

build/Makefile.simple
```diff
@@ -9,6 +9,7 @@ ifndef BINDIR
 $(error build/Makefile.vars should set BINDIR)
 endif
 
+SHARED_SOURCES ?=
```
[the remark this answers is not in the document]
ltratt: Ok, I can buy the documentation reason -- I mostly wanted to check this wasn't a mistake!

hybrid/ddc_compartment_switching/ddc_compartment_switching_nok.c
drpdr: Fix in 39f3384.

ltratt:
Am I correct in thinking that we're not currently testing the success / failure of running programs in CI?

drpdr:
> Can we at least test things on _one_ of the instances? It's better to test success/failure on just (say) Morello than not test it on either Morello nor RISC-V.

Oh yes, I'll work on this too. In this PR?

ltratt:
Yes, we really need to test things at least a bit in this PR.

drpdr:
> Yes, we really need to test things at least a bit in this PR.

Tests in 61207a7.

drpdr:
> From buildbot:
> 
> ```
> Running tests took 0:00:06.989786
> ===> DONE
> Total execution time: 0:01:11.237422
> ```
> 
> Did that actually run any tests? I thought they printed something to `stdout`?

I thought we did not want anything to `stdout`, so I only check [here](https://github.com/capablevms/cheri-examples/blob/61207a73bab5c259e6a3455e3ba2391f2fcbe917/tests/run_tests.sh#L44) and `exit` accordingly.
I can add a message if you wish, though.

drpdr:
> > I thought we did not want anything to stdout
> 
> That's correct: I just had a vague memory that we printed out which tests we were executing? If I'm wrong, no problem, I just wanted to check that the things we think should be run are being run!

There is the `ssh` line printed because we do not suppress anything. We can add an `echo` at the beginning if you think it is better/more clear.

ltratt:
Let's not worry about that. I think the last comment for you to deal with is https://github.com/capablevms/cheri-examples/pull/44#commitcomment-57661506?

PR #49, opened 2021-11-16
Compartment bind and sentry

0152la:
Add two examples showcasing compartment features

The example in d397b1e7cf4be9939ad87a1bf252073f725419b6 showcases how a restricted PCC, after being unbound to
perform a sentry call, remains unbound, allowing a compartment to access
code it should not be privy to. Particularly, after calling `comp_fun_c` via
a sentry at `shared.s:110`, we are able to call `comp_fun_c` directly (e.g.,
`shared.s:89`).

The example in fb03315c91408d0a04761db11c8c534e636c6560 showcases two particular actions to help compartmentalization:
     - reducing the bounds of the PCC (shared.s:82)
     - calling a sentry from within a bounded compartment (shared.s:114)
In addition to the above, we note that an implicit PCC unbounding is
performed when a sentry is called in hybrid mode. The PCC is then **not**
implicitly rebound. This example uses a trampoline function `comp_fun_tr`
which does this rebounding.

### DIFF
--- a/hybrid/compartment_examples/call_sentry_and_unbound_pcc/main.c
+++ b/hybrid/compartment_examples/call_sentry_and_unbound_pcc/main.c
@@ -1,0 +1,55 @@
+/***
+ * This example showcases how a restricted PCC, after being unbound to
+ * perform a sentry call, remains unbound, allowing a compartment to access
+ * code it should not be privy to. Particularly, after calling `comp_fun_c` via
+ * a sentry at `shared.s:110`, we are able to call `comp_fun_c` directly (e.g.,
+ * `shared.s:89`).
+ * ***/
+
+#include "../../../include/common.h"
+#include "../../include/utils.h"
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+// Trampoline function to ensure PCC is rebounded after we return from a sentry
+// call; this assembly function also includes the sentry containing the
+// function to be called from the container
+extern int comp_fun();
+// The function connected to the assembly trampoline
+extern int switch_compartment(void *stack, size_t size, void* __capability fn_call_start, void* __capability pcc);
+
+// Function to be called via a sentry from within a restricted compartment. As
+// the `stk` variable is within DDC, it can write to it.
+void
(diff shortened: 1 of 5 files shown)


ltratt:
@0152la You should assign someone (probably me) to the PR under "assignees" to the right.

ltratt:
This might not be (fully) possible, but we're now growing lots of variants of `shared.s` and it's getting a bit difficult to see what the difference between each is. Might it be possible to merge some of the variants together somewhat?

0152la:
That is a concern, as well as copies of `Makefile.morello-hybrid`. I can try to think of a potentially better approach. However, I think since each example is fully contained by itself, within a separate folder, it makes the distinction clear? Merging them for specific examples and having one zombie example might make it less clear in the big picture, unless we do something very smart.

ltratt:
Yes, I'm not entirely sure what the right answer is either. I'm partly thinking that at the moment it's very hard to see what different examples do when they only vary by 1 or 2 lines in ~200LoC. It might not be something we can sensibly fix in this PR, but it's worth having a think about it IMHO.

hybrid/compartment_examples/call_sentry_and_unbound_pcc/main.c
```diff
@@ -0,0 +1,55 @@
+/***
+ * This example showcases how a restricted PCC, after being unbound to
```
ltratt, line 2: Silly question: what does "unbound" mean?

0152la: I mean it here as changing the "bounds" fields within the capability to (presumably, have not checked) cover the whole memory space, rather than the limited space we constrain it to. Which I believe happens as a side-effect of unsealing, overwriting our set `PCC` via a jump instruction.

ltratt: Let me try rephrasing what you've said and see if it makes sense.

If I call a sealed function `f` via a capability `C`, the `PCC` will be set to `C`'s bounds during execution. After `f` returns, the `PCC` remains set to whatever `C`'s bounds were.

Is that correct?

0152la: I have confirmed this and you are correct; the `PCC` is set to the bounds set within the sentry. Which now puts the issue that when we return from `C` code back to assembly, we can't execute anymore unless the PCC contained within the sentry covers that code...

0152la:
After further discussion, the examples here were written with a misunderstanding of how calling a sentry interacts with updating the PCC. They will be adapted, correcting this misunderstanding.

PR #50, opened 2021-11-19
Compartment call sentry

0152la:
(this is an improvement over #49)

These examples aim to do two things:
* restrict the program capability counter (`PCC`)
* call a function outside PCC bounds via a sealed capability entry
  (sentry)

Example `restrict_pcc` shows that we successfully restrict the `PCC`,
and that calling a function outside the `PCC` bounds fails, with an
expected `In-address space security exception`.

Example `call_sentry` shows how to call a function outside of restricted
`PCC` bounds via a sentry, as well as one (not fully secure) choice of
how to transition between a compartment (i.e., restricted permissions)
and a state with full permissions.

### DIFF
--- a/hybrid/compartment_examples/call_sentry/main.c
+++ b/hybrid/compartment_examples/call_sentry/main.c
@@ -1,0 +1,92 @@
+/***
+ * This example showcases two particular actions to help compartmentalization:
+ *      - reducing the bounds of the PCC (shared.s:82)
+ *      - calling a sentry from within a bounded compartment (shared.s:114)
+ *
+ * One difficulty, and capability leak, is how to return from the sentry
+ * function back to the caller with correct capabilities (i.e., PCC). In this
+ * example, we use `clr`, and manually return to it (main.c:49). However,
+ * having the `clr` available to the sentry function means it could access
+ * capabilities it should not be privy to (e.g., if the bounds of the sentry
+ * are set to only cover the specific function). In this example, this is
+ * further aggravated by the jump between assembly and `C`: the `C` function is
+ * compiled to assembly which makes use of `lr` (which stores pointers, not
+ * capabilities), meaning we would, by default, return to a context in which we
+ * have no execution permissions (more specifically, the `PCC` is set to cover
+ * the callee, and when we return to the caller, the `PCC` is not affected due
+ * to returning to a pointer).
+ ***/
+
+#include "../../../include/common.h"
+#include "../../include/utils.h"
+
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+// Trampoline function to restore caller's PCC after we return from a sentry
+// call; this assembly function also includes the sentry containing the
+// function to be called from the container
+extern int comp_fun_tr();
+// The function connected to the assembly trampoline
+extern int switch_compartment(void *stack, size_t size, void *__capability fn_call_start,
+							  void *__capability pcc);
+
+// Function to be called via a sentry from within a restricted compartment. As
+// the `stk` variable is within DDC, it can write to it.
+void comp_fun_c(uint8_t *stk)
+{
+	stk[1800] = 80;
+
+	// The `add` instruction is required to correctly reset the `sp`, which is
+	// modified by the assembled version of this function; this might be
+	// compiler-specific. In this example, `sp` is used to store information to
+	// reset the environment, therefore we must ensure it is reset correctly
+	// upon return.
+	//
+	// The `ret` instruction produced by
+	// default only uses `lr`, instead of `clr`; using the `lr` means we would
+	// return with a PCC bound to this function, and not be able to execute
+	// within the caller. Using `clr` ensures the PCC is reset correctly.
+	//
+	// Overall, this example shows it is not a good idea to call C functions
+	// via sentries in hybrid mode.
+	//
+	// NOTE: As the `clr` remains in scope during the execution of this
+	// function, it essentially leaks the executable capability. The current
+	// design, and the interfacing between `C` and assembly requires this to be
+	// available in this function
+	asm("add sp, sp, #0x10; ret clr");
+}
+
+int main()
+{
+	uint8_t *comp_mem = malloc(5000);
+	size_t comp_size = 2000;
+
+	// Create a capability which we will use to tightly bound the PCC for the
+	// compartment
+	void *__capability call_cap = (void *__capability) comp_fun_tr;
+	call_cap = cheri_bounds_set(call_cap, comp_size);
+
+	// Derive a sentry from the PCC to allow calling a function outside PCC
+	// bounds
+	void *__capability comp_fun_c_sentry = cheri_pcc_get();
+	comp_fun_c_sentry = cheri_address_set(comp_fun_c_sentry, (unsigned long) &comp_fun_c);
+
+	// Set the bounds on the sentry, which restricts the PCC when executing the
+	// function. The value `40` here was determined by disassembling
+	// `comp_fun_c`.
+	comp_fun_c_sentry = cheri_bounds_set(comp_fun_c_sentry, 40);
+
+	switch_compartment(comp_mem, comp_size, call_cap, comp_fun_c_sentry);
+	//
+	// Check the compartment function has been executed.
+	assert(comp_mem[1800] == 80);
+	return 0;
+}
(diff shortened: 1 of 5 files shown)


hybrid/compartment_examples/call_sentry/main.c
```diff
@@ -0,0 +1,88 @@
...
+ * This example showcases two particular actions to help compartmentalization:
+ *      - reducing the bounds of the PCC (shared.s:82)
+ *      - calling a sentry from within a bounded compartment (shared.s:114)
+ *
+ * One difficulty, and capability leak, is how to return from the sentry
+ * function back to the caller with correct capabilities (i.e., PCC). In this
+ * example, we use `clr`, and manually return to it (main.c:49). However,
+ * having the `clr` available to the sentry function means it could access
+ * capabilities it should not be privy to (e.g., if the bounds of the sentry
+ * are set to only cover the specific function. In this example, this is
```
ltratt, line 11: Missing ")" before "."?

hybrid/compartment_examples/call_sentry/main.c
```diff
@@ -0,0 +1,88 @@
...
+ * further aggravated by the jump between assembly and `C`: the `C` function is
+ * compiled to an assembly which makes use of `lr` (which stores pointers, not
```
ltratt, line 13: s/an assembly/assembly/

hybrid/compartment_examples/call_sentry/main.c
```diff
@@ -0,0 +1,88 @@
...
+ *      - reducing the bounds of the PCC (shared.s:82)
+ *      - calling a sentry from within a bounded compartment (shared.s:114)
+ *
+ * One difficulty, and capability leak, is how to return from the sentry
+ * function back to the caller with correct capabilities (i.e., PCC). In this
+ * example, we use `clr`, and manually return to it (main.c:49). However,
+ * having the `clr` available to the sentry function means it could access
+ * capabilities it should not be privy to (e.g., if the bounds of the sentry
+ * are set to only cover the specific function. In this example, this is
+ * further aggravated by the jump between assembly and `C`: the `C` function is
```
ltratt, line 12: I'm not sure how the jump between assembly and C aggravates things? Making use of `lr` isn't a (capability) problem as such. I agree that it's slightly annoying/awkward to have to transition from a capability-aware thing (in assembly) and a non-capability-aware thing (in C), but I'm not sure I can see this adding security issues?

hybrid/compartment_examples/call_sentry/main.c
```diff
@@ -0,0 +1,88 @@
...
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+// Trampoline function to ensure PCC is rebounded after we return from a sentry
```
ltratt, line 32: I think we might want to drop "unbound" and "rebound" as their intuitive meaning is easily confused with what's actually going on. It might be better to say something like "Trampoline function to restore the caller's PCC after a sentry returns" ?

hybrid/compartment_examples/call_sentry/main.c
```diff
@@ -0,0 +1,88 @@
...
+    // The `ret` instruction produced by
+    // default only uses `lr`, instead of `clr`; using the `lr` means we would
+    // return with a PCC bound to this function, and not be able to execute
+    // within the caller. Using `clr` ensures the PCC is reset correctly.
+    //
+    // NOTE: As the `clr` remains in scope during the execution of this
+    // function, it essentially leaks the executable capability. The current
+    // design, and the interfacing between `C` and assembly requires this to be
+    // available in this function
+    asm("add sp, sp, #0x10; ret clr");
```
ltratt: Hmm, this does highlight something we touched upon when we talked in person: it's quite difficult for a C function to reliably be compiled in hybrid mode but operate (sometimes) in sort-of purecap mode (at least when it returns). The fragility of the `add 10` thing is probably the killer example for that!

Perhaps we should say that we only ever compile C code in fully-hybrid or fully-purecap mode? Then the assembly compartment switcher can do all the black magic of converting between the two modes (i.e. the sentry that I call would be an assembly stub which then calls a normal-hybrid-mode C function; the C function returns via the `lr` to the assembly stub; and then the assembly stub returns via the `clr` to the sentry-caller).

hybrid/compartment_examples/call_sentry/main.c
0152la: Addressed in c9a02f2c81cc1b627dd38f71d2ab44bd3113fa16.

hybrid/compartment_examples/call_sentry/main.c
0152la: Addressed in c9a02f2c81cc1b627dd38f71d2ab44bd3113fa16.

hybrid/compartment_examples/call_sentry/main.c
0152la: Addressed in c9a02f2c81cc1b627dd38f71d2ab44bd3113fa16.

hybrid/compartment_examples/call_sentry/main.c
0152la: The core of the issue still remains in hybrid mode: if you have a sentry to a C function, you can't fully utilize it (as far as I understand it via these examples) as the PCC will not be reset, and you will return to unexecutable code. There is the possibility of dissecting the sentry, and potentially deriving a PCC to cover both the function to be executed and the return location, but that would require the PCC be set beyond the bounds initially defined in the sentry.

hybrid/compartment_examples/call_sentry/main.c
0152la: I didn't mean "further aggravat[ing]" the security stuff, more that horrendous `add sp` instruction. I can reword it if it sounds out of place.

hybrid/compartment_examples/call_sentry/main.c
ltratt: I think we're more-or-less in agreement: what we're both saying is that if C code is compiled in hybrid mode calling it in "sotr of purecap mode" (i.e. via a sentry) leads to all sorts of weird contortions.

To put it another way: in hybrid mode we should, I think, say "C functions should not be called via a sentry"?

0152la: Right, but if we say that, then the idea of "mutually distrusting compartments" goes out the window, I believe.

ltratt: I don't think so: it means you have to go via assembly. The problem isn't that compartmentalisation is impossible, it's just that in hybrid mode C functions aren't compiled in a way that you can use them directly with sentries.

0152la: Ah right, I misread. Yes, I do believe these issues would not crop up in a pure assembly example. And, for completion, we can use C function sentries, just not to the full extent of their security capabilities (i.e., in our design, we need to leak a more permissive PCC capability).

ltratt: > i.e., in our design, we need to leak a more permissive PCC capability

Why's that?

0152la: So that the C function can return to the caller (as we discussed above; this is illustrated in fb03315c91408d0a04761db11c8c534e636c6560).

ltratt: Right, but that's not really "our design" as such, I think, but an inevitable consequence of the C code being compiled in hybrid mode?

0152la: I'm leaving the door open to some possible API/sequence of assembly instructions that might be able to do this, but we are unaware of.

ltratt: I think we should close the door on it and this PR nicely shows why :) It's clearly *possible* to put inline assembly in that can return from a sentry, but equally clear that it's not portable (at the most obvious: RISC-V vs. Morello; but it probably isn't even stable across compiler versions etc.).

0152la: Fair enough. I wanted to just specify it for historic purposes, in case anyone stumbles upon it.

ltratt: I think it would be fine to say something like "The only way to avoid this problem would be for the hybrid mode C compiler to do XYZ" which will help the future reader more.

0152la: I chose the coward's way out in the example and just went ahead with the "don't do C sentry calls in hybrid" angle.

ltratt: I don't think it's the coward's way out: I think it's the right way :)

ltratt:
Please squash.

0152la:
Squashed.

PR #52, opened 2021-11-30
Compartment switch via memory stored data

0152la:
This example builds on previous ones, particularly improving the
`switch_compartment` (`shared.s:17`) function. This is a step towards
inter-compartment transition without the need of a more-priveleged entity
controlling the switch (but needing to set everything up before control is
relinquished to a compartment).

Some specific improvements over previous examples:
- the compartment switching function now reads information from memory
rather than from the stack. This allows compartments to call this function
from a constrained environment, without leaking capabilities beyond a few
prepared ones;
- we showcase the use of the compartment ID (`CID_EL0`) register to control
  which compartment we are switching into. This is more of an example, as it
  would entail compartments are aware of the ID of the callee.

### DIFF
--- a/hybrid/compartment_examples/comp_setup/main.c
+++ b/hybrid/compartment_examples/comp_setup/main.c
@@ -1,0 +1,136 @@
+/* This example builds on previous ones, particularly improving the
+ * `switch_compartment` (`shared.s:17`) function. This is a step towards
+ * inter-compartment transition without the need of a more-privileged entity
+ * controlling the switch (but needing to set everything up before control is
+ * relinquished to a compartment).
+ *
+ * Some specific improvements over previous examples:
+ * - the compartment switching function now reads information from memory
+ * rather than from the stack. This allows compartments to call this function
+ * from a constrained environment, without leaking capabilities beyond a few
+ * prepared ones;
+ * - we showcase the use of the compartment ID (`CID_EL0`) register to control
+ *   which compartment we are switching into. This is more of an example, as it
+ *   would entail compartments are aware of the ID of the callee.
+ */
+
+#include "../../../include/common.h"
+#include "../../include/utils.h"
+
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+/*******************************************************************************
+ * Assembly Functions & Consts
+ ******************************************************************************/
+
+extern int switch_compartment();
+extern void comp_f_fn();
+
+/*******************************************************************************
+ * Types & Consts
+ ******************************************************************************/
+
+const size_t comp_stack_size = 2000;
+const size_t total_comp_size = 5000;
+size_t id = 0;
+
+/* Abstract representation of a compartment. Within 80 bytes we represent:
+ * - an id (8B)
+ * - address to the start of the compartment's stack (8B)
+ *   the size of the stack (8B)
+ * - address to the start of the compartment's heap (8B)
+ *   the size of the heap (8B)
+ * - alignment padding (8B)
+ * - the ddc corresponding to the compartment (16B)
+ * - the function within the compartment to be executed upon switching (16B)
+ */
+struct comp
+{
+	size_t id;
+	void *stack_addr;
+	size_t stack_len;
+	void *heap_addr;
+	size_t heap_len;
+	void *__capability ddc;
+	void *__capability comp_fn;
+};
+
+// ASM offsets, included here for validation
+#include "main.h"
+
+static_assert(COMP_SIZE == sizeof(struct comp), "Invalid `COMP_SIZE` provided");
+static_assert(COMP_OFFSET_STK_ADDR == offsetof(struct comp, stack_addr),
+			  "Invalid `COMP_OFFSET_STK_ADDR` provided.");
+static_assert(COMP_OFFSET_STK_LEN == offsetof(struct comp, stack_len),
+			  "Invalid `COMP_OFFSET_STK_LEN` provided.");
+static_assert(COMP_OFFSET_DDC == offsetof(struct comp, ddc), "Invalid `COMP_OFFSET_DDC` provided.");
+static_assert(COMP_OFFSET_PCC == offsetof(struct comp, comp_fn),
+			  "Invalid `COMP_OFFSET_PCC` provided.");
+
+struct comp comps[COMP_COUNT];
+
+/*******************************************************************************
+ * Privileged Functions
+ ******************************************************************************/
+
+/* This function stores the compartment data in memory, so that it can be
+ * accessed by the compartments as needed.
+ */
+void executive_switch(struct comp c)
+{
+	void *comps_addr = &comps;
+	asm("mov x19, %0\n\t"
+		"mov x0, #0\n\t"
+		"msr CID_EL0, c0"
+		:
+		: "r"(comps_addr));
+
+	switch_compartment();
+}
+
+void add_comp(uint8_t *_start_addr, void (*_comp_fn)())
+{
+	assert(id < COMP_COUNT);
+
+	struct comp new_comp;
+	new_comp.id = id;
+
+	new_comp.stack_addr = (void *) _start_addr;
+	new_comp.stack_len = comp_stack_size;
+	new_comp.heap_addr = (void *) (_start_addr + comp_stack_size);
+	new_comp.heap_len = total_comp_size - comp_stack_size;
+
+	void *__capability comp_ddc = (void *__capability) _start_addr;
+	comp_ddc = cheri_bounds_set(comp_ddc, total_comp_size);
+	new_comp.ddc = comp_ddc;
+
+	void *__capability comp_fn = (void *__capability) _comp_fn;
+	comp_fn = cheri_bounds_set(comp_fn, 40);
+	new_comp.comp_fn = comp_fn;
+
+	comps[id] = new_comp;
+	++id;
+}
+
+/*******************************************************************************
+ * Main
+ ******************************************************************************/
+
+int main()
+{
+	uint8_t *comp_f = malloc(total_comp_size);
+	add_comp(comp_f, comp_f_fn);
+
+	executive_switch(comps[0]);
+
+	// Check compartment did indeed execute
+	assert(comp_f[4000] == 42);
+}
--- a/hybrid/compartment_examples/comp_setup/shared.S
+++ b/hybrid/compartment_examples/comp_setup/shared.S
@@ -1,0 +1,142 @@
+// Copyright (c) 2021 The CapableVMs "CHERI Examples" Contributors.
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+#include "main.h"
+
+.global comp_f_fn
+
+.text
+.balign 4
+
+/* This function has been updated with a few features:
+ * - it identifies which compartment to switch to based on the value within the
+ * `CID_EL0` register;
+ * - it reads compartment data from memory, rather than it having to be passed
+ * to it via registers - this includes DDC and PCC of compartment to switch to.
+ */
+.global switch_compartment
+.type switch_compartment, "function"
+switch_compartment:
+
+    // Get compartment to switch to data
+    mrs       c10, CID_EL0
+    mov       x11, #COMP_SIZE
+    mul       x10, x10, x11
+
+    // Setup SP
+    mov       x12, sp
+    add       x11, x10, #COMP_OFFSET_STK_ADDR
+    ldr       x11, [x19, x11]
+    add       x13, x10, #COMP_OFFSET_STK_LEN
+    ldr       x13, [x19, x13]
+    add       sp, x11, x13
+
+    // Derive a new clr to restore PCC, and store it.
+    cvtp      c11, lr
+
+    // Save old DDC, old SP, old LR on stack
+    mrs       c10, DDC
+    stp       c10, c11, [sp, #-48]!
+    str       x12, [sp, #32]
+
+    // Stack layout at this point:
+    //
+    //     `stack + size` -> ________________________
+    //            sp + 40 -> [  <alignment pad>  ]   ^
+    //            sp + 32 -> [      old SP       ]   |
+    //            sp + 24 -> [ old CLR (hi64)    ]   |
+    //            sp + 16 -> [ old CLR (lo64)    ]   |
+    //            sp +  8 -> [ old DDC (high 64) ]   | DDC bounds
+    //            sp +  0 -> [ old DDC (low 64)  ]   |
+    //                                 :             :
+    //            `stack` -> ________________________v
+
+    // Load PCC, including function we are jumping to within compartment
+    add       x11, x10, #COMP_OFFSET_PCC
+    ldr       c0, [x19, x11]
+
+    // Load DDC
+    add       x11, x10, #COMP_OFFSET_DDC
+    ldr       c11, [x19, x11]
+    msr       DDC, c11
+
+    bl        clean+4
+
+    // Jump to the function within the compartment we are switching to
+    blr       c0
+
+    // Clean capabilities left in the return value.
+    mov       w0, w0
+    bl        clean
+
+    // Restore the caller's context and compartment.
+    ldp       c10, clr, [sp]
+    ldr       x12, [sp, #32]
+    msr       DDC, c10
+    mov       x10, #0
+    mov       sp, x12
+
+    ret       clr
+
+/* The function in this compartment just writes to some memory within its
+ * bounds, to ensure it is properly called.
+ */
+.type comp_f_fn, "function"
+comp_f_fn:
+    mrs c10, DDC
+    mov x11, 42
+    str x11, [x10, #4000]
+
+    ret clr
+
+
+    // Inner helper for cleaning capabilities from registers, either side of an
+    // AAPCS64 function call where some level of distrust exists between caller
+    // and callee.
+    //
+    // Depending on the trust model, this might not be required, but the process
+    // is included here for demonstration purposes. Note that if data needs to
+    // be scrubbed as well as capabilities, then NEON registers also need to be
+    // cleaned.
+    //
+    // Callers should enter at an appropriate offset so that live registers
+    // holding arguments and return values (c0-c7) are preserved.
+clean:
+    mov x0, #0
+    mov x1, #0
+    mov x2, #0
+    mov x3, #0
+    mov x4, #0
+    mov x5, #0
+    mov x6, #0
+    mov x7, #0
+    mov x8, #0
+    mov x9, #0
+    mov x10, #0
+    mov x11, #0
+    mov x12, #0
+    mov x13, #0
+    mov x14, #0
+    mov x15, #0
+    mov x16, #0
+    mov x17, #0
+    // x18 is the "platform register" (for some platforms). If so, it needs to
+    // be preserved, but here we assume that only the lower 64 bits are
+    // required.
+    mov x18, x18
+    // x19-x29 are callee-saved, but only the lower 64 bits.
+    mov x19, x19
+    mov x20, x20
+    mov x21, x21
+    mov x22, x22
+    mov x23, x23
+    mov x24, x24
+    mov x25, x25
+    mov x26, x26
+    mov x27, x27
+    mov x28, x28
+    mov x29, x29  // FP
+    // We need LR (x30) to return. The call to this helper already cleaned it.
+    // Don't replace SP; this needs special handling by the caller anyway.
+    ret
+
--- a/hybrid/compartment_examples/comp_setup/main.h
+++ b/hybrid/compartment_examples/comp_setup/main.h
@@ -1,0 +1,6 @@
+#define COMP_COUNT 1
+#define COMP_SIZE 80
+#define COMP_OFFSET_STK_ADDR 8
+#define COMP_OFFSET_STK_LEN 16
+#define COMP_OFFSET_DDC 48
+#define COMP_OFFSET_PCC 64


0152la:
One thing that I left in is if we can use `#define`d constants in assembly. I saw the Android Morello example does so, but I was unable to replicate this in my test. Examples of this can be seen [here](https://github.com/capablevms/cheri-examples/pull/52/files#diff-ed2cd41376ca9f400f68119c052d2b46bd0324d982543ceb5c645813d5c1c75fR27).

ltratt:
I must admit, I haven't quite understood the high-level intent of this PR: in particular I'm not quite sure what "the compartment switching function now reads information from memory rather than from the stack. This allows compartments to call this function from a constrained environment, without leaking capabilities beyond a few prepared ones;" means. You might need to give me a simpler explanation!

0152la:
I think the main reason is to ensure what I'm doing is better explained, so you not understanding goes very well with that intent. 

In our `ddc_switching` [1] and the `call_sentry` [2] examples, we pass information regarding the compartment switch (i.e., DDC, PCC, function to call) registers. That means (a) we have a limited amount of registers to store that information, and (b) we need to be extremely careful to clear those registers to not leak capabilities.

Now, as we are trying to perform an inter-compartment switch, generally, we can assume the compartment should have access to any other compartment defined in the system. This directly means passing all these things via registers is rather intractable. Thus, moving to memory, and ensuring everything is stored/loaded correctly is one thing this example shows. Furthermore (not shown in this example), upon issuing a compartment switch request, we can further restrict the DDC to the area of memory we know is required for switching, since it's all bundled nicely together.

Further, this example is more structured than the previous ones, where the function would expect certain information in respective registers. Here, the setup is done exclusively in C code (via the `comps` array), and then the assembly uses that information appropriately, making it (hopefully) more understandable. This will be more obvious how to use once a second (and potentially more) compartments come into play.

And finally, I want to have this PR so someone else can look at it and say it seems reasonable, and I've not missed something obvious or used something inappropriately.

[1] https://github.com/capablevms/cheri-examples/blob/master/hybrid/ddc_compartment_switching/ddc_compartment_switching.c
[2] https://github.com/capablevms/cheri-examples/blob/master/hybrid/compartment_examples/call_sentry/main.c

jacobbramley:
> One thing that I left in is if we can use `#define`d constants in assembly. I saw the Android Morello example does so, but I was unable to replicate this in my test. Examples of this can be seen [here](https://github.com/capablevms/cheri-examples/pull/52/files#diff-ed2cd41376ca9f400f68119c052d2b46bd0324d982543ceb5c645813d5c1c75fR27).

You can, if you run the assembly file through the C preprocessor. That is common practice, and if you name the file ".S" instead of ".s", [it will happen automatically](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html). Some projects have #-directive-only headers designed for inclusion into both ".c" and ".S" files.

hybrid/compartment_examples/comp_setup/main.c
```diff
@@ -0,0 +1,118 @@
+/* This example builds on previous ones, particularly improving the
+ * `switch_compartment` (`shared.s:17`) function. This is a step towards
+ * inter-compartment transition without the need of a more-priveleged entity
```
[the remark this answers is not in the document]
0152la: Fixed in https://github.com/capablevms/cheri-examples/pull/52/commits/a2ff4f3c6821dff8e036707aea754c06e79bdafb.

0152la:
Thanks for the tip, Jacob. Updated in https://github.com/capablevms/cheri-examples/pull/52/commits/a2ff4f3c6821dff8e036707aea754c06e79bdafb.

ltratt:
I wonder if this might be a good place for some ASCII art to sketch out the compartment(s) / compartment switcher and their relationship? [A bit like the comment in https://github.com/capablevms/cheri-examples/pull/52/files#diff-9765feb1c5a50fb266ee211ee57158a3aac1d06f9cad5cb4866241d1d900d020R43.] There is software which apparently helps do this, although I can't say I've used it!

0152la:
Here's further clarification on the functionality of this example. The main takeaway is that there is *no* switcher entity per se, as this is an intermediate example towards inter-compartment switching.

The `C` code stores in memory data needed to perform a compartment switch (including the DDC and PCC of the compartment to switch to). Then, the `switch_compartment` function loads that data and performs the switch. The compartment we are switching to is defined by register `CID_EL0` (there is no special reason to use this register particularly, other than to illustrate we can use it, and that it is a possible use case for this register). In this example, there is a single compartment, represented by `comp_f_fn` [1] (this stores a value to memory just to validate the example). The idea would be that `comp_f_fn` will be able to then call `switch_compartment` itself, in a subsequent example.

There is no need for a separate higher-privileged compartment switcher, ss all necessary information to perform compartment switching is pre-composed and stored in memory.

[1] https://github.com/capablevms/cheri-examples/pull/52/files#diff-9765feb1c5a50fb266ee211ee57158a3aac1d06f9cad5cb4866241d1d900d020R85

ltratt:
> The C code stores in memory data needed to perform a compartment switch

Can we expand upon this a bit: what's being stored and where? For example you might say something like "We allocate room for a new compartment stack + heap + room for parameters to be passed from the outside" or similar.

0152la:
We abstract a compartment as a `struct comp` [1], and we have an array `comps` of such `struct comp`s. The struct contains an id (which currently has not much of a functional use, but could be used, for example, to cross-check that the compartment we are loading via the `CID_EL0` matches the stored compartment), a capability meant to be the DDC of the compartment, and a capability meant to be used as the PCC, pointing to the entry function (we decided that for these examples, we'd limit each compartment to contain a single function). Adding a compartment is done via `add_comp` [2].

[1] https://github.com/capablevms/cheri-examples/pull/52/files#diff-ba1ccd56981d65877ec12ee23b66c30718e755164c316685032972fae92a7879R46
[2]  https://github.com/capablevms/cheri-examples/pull/52/files#diff-ba1ccd56981d65877ec12ee23b66c30718e755164c316685032972fae92a7879R79

ltratt:
Ah, so I *think* that we might have had a slight disconnect between what "setup a compartment means". I thought it meant:

1. Create the tracking info for the compartment (what is currently the `comp` struct) so that a (future) switcher can keep track of different compartments.
2. Allocate the memory (stack / heap) for it.

I *think* this PR just does the first of those two steps?

0152la:
I would argue it is (at least on attempt of) doing both. The problem is with the stack, but that's a detail.

So point 1. Again, the main idea is that there is no "switcher" entity to perform switching. It is a process readily available at any point in the system. Now, we could call the process a switcher, but the process doesn't "hold control" over any data: the data it uses is this "tracking data" (i.e., the `comps` array). What I personally understand by a "switcher" is an implicit privilege escalation: a compartment, with its own restricted view of the system, cannot transition to another compartment (we can only *restrict* capability bounds, not modify them arbitrarily; as two compartments would, presumably, cover different bounds for both the PCC/DDC, a compartment can then not transition by itself to other bounds unless either we escalate to something that can (e.g., a switcher), or we use pre-constructed capabilities (e.g., this example)). 

Regarding point 2, I do attempt to do this. I allocate 5k bytes [1] for our compartment, and restrict the DDC to that area of memory [2]. I am unsure how to distinguish between stack and heap within this 5k block of memory, but I did attempt to copy what Jacob did [3], but wouldn't be able to explain if it's correct or not. However, the idea is that the DDC we store for the respective compartment points to the start of the memory allocated for that compartment, and we know it's a set 5k bytes, with 2k bytes of stack (arbitrary numbers chosen).

[1] https://github.com/capablevms/cheri-examples/pull/52/files#diff-ba1ccd56981d65877ec12ee23b66c30718e755164c316685032972fae92a7879R107
[2] https://github.com/capablevms/cheri-examples/pull/52/files#diff-ba1ccd56981d65877ec12ee23b66c30718e755164c316685032972fae92a7879R87
[3] https://github.com/capablevms/cheri-examples/pull/52/files#diff-9765feb1c5a50fb266ee211ee57158a3aac1d06f9cad5cb4866241d1d900d020R33

ltratt:
Aha, I'm with you.

I'm going to suggest that we tweak this PR a little bit to save us some future pain. I think the struct (which are typically capitalised in modern C, but that's nit picking on my behalf!) should be something like:

```c
typedef CompartmentId size_t;

struct Compartment {
  id CompartmentId;
  void *stack;
  size_t stack_len;
  void *heap;
  size_t heap_len;
  void* __capability ddc;
  void *() __capability entry_point;
}
```

In other words, we need to have specific pointers to the stack and heap, and store their lengths explicitly (rather than as constants). This way we make a bit clearer that these things might change (e.g. you might want greater heap) and we also make clear that the bounds of the DDC must cover the stack and heap but (because of bounds approximation) might cover more or less of the stack and heap than we actually want. There's an implicit contract that `ddc >= address(stack) && ddc >= address(heap) && bounds(ddc) >= stack_len + heap_len` that's probably worth having as an `assert` in one or more places.

I'm not sure if we need the `id` as at the moment it's implicit in the `Compartment`s offset in the array. If we don't need it, I'd be inclined to drop it, as identifiers can cause problems in the long run (e.g. if we delete a compartment, can an Id be reused?).

I'm neutral as to whether `comp_func` should be part of the `Compartment` struct or not. My first thought was "we don't need it" but on reflection I suppose a fairly general model of compartments will expose 1 or more functions to the outside world: so I guess having a single function pointer `entry_point` (or whatever we might want to call it/them) is reasonable. I'm not sure what its type signature should be, but we should probably make clear at least that it's a function in the struct.

0152la:
Alright. Beyond these functional changes, does the functionality of the example now make sense? And should I expand on any of the comments in the test based on this discussion?

Regarding the functional changes, I would keep it to a minimum, to reduce memory footprint. The DDC `bounds` field should contain the address of the stack (lower bound) and heap (upper bound), and we'd need just the length of the stack to identify where they meet. Thus, this would save 16*3 bytes of memory (I believe structs are aligned based on the largest member, in our case the 16 byte capabilities).

Also could you explain what `ddc >= address(stack)` means? As far as I understand it, the `DDC`'s "value" is just the address it points to, which in this case should be the address where the compartment starts.

ltratt:
> Beyond these functional changes, does the functionality of the example now make sense? 

I think so.

> Regarding the functional changes, I would keep it to a minimum, to reduce memory footprint

The size of the `Compartment` struct is so tiny compared to the actual compartment that I don't think it's worth trying to save a few bytes, especially because I bet you that saving will bite us when someone forgets that the DDC's bounds are approximate.

>  Also could you explain what ddc >= address(stack) means?

I was just being abstract as to whether the heap or stack comes first in the compartment. If we (say) always have the stack first then the assert can be changed accordingly.

0152la:
I've updated the compartment representation. The insufficiencies at this stage in time (more for documentation, than for "we should fix these right now"):

* the compartment heap is unusable - this would require a custom `malloc`, or writing our own small `asm` snippet to handle this; however, the heap is modelled in the code
* the DDC is not restricted when calling `switch_compartment`; in the current design, I can't think of a way to not leak this to the compartments, but we could do this via Morello's executive/restricted feature (which would mean entrenching Morello-specificity even further). However, there is a conversation to be had whether leaking limited compartment info is bad - this is information needed for the compartments to communicate (i.e., transition).

ltratt:
> However, there is a conversation to be had whether leaking limited compartment info is bad - this is information needed for the compartments to communicate (i.e., transition).

Good point. My gut reaction is that we don't want to leak this info except when we explicitly choose to do so -- but we don't have to go that correct in this PR. [My first thought is that I was wrong above: we do need compartment Ids so that compartments can reference each other without having to be able to access any part of memory to do with the compartment directly. That suggests that compartment IDs should be sealed capabilities and that deleting a compartment might require doing a full revocation of that sealed capability. That's way too much for this PR though!]

hybrid/compartment_examples/comp_setup/main.c
```diff
@@ -43,13 +42,23 @@ const size_t total_comp_size = 5000;
 size_t id = 0;
 const size_t comp_cnt = 1;
 
-// For each compartment, we will store to memory:
-// * an id
-// * the ddc corresponding to the compartment
-// * the function within the compartment to be executed upon switching
+/* Abstract representation of a compartment. Within 80 bytes we represent:
```
ltratt, line 45: I'm not sure we want to put the sizes in here since they're not really guaranteed. FWIW, I think the size of the struct is fine and we don't have to worry about it!

hybrid/compartment_examples/comp_setup/shared.S
```diff
@@ -2,9 +2,11 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 #define COMP_COUNT 1
-#define COMP_SIZE 48
-#define COMP_OFFSET_DDC 16
-#define COMP_OFFSET_PCC 32
+#define COMP_SIZE 80
```
ltratt, line 5: Is it possible to import the header file from the `.S` file and use `sizeof`? [This might be a very stupid question that everyone knows is impossible!]

hybrid/compartment_examples/comp_setup/main.c
0152la: The sizes are more for me so I can easily calculate the offsets in `shared.S` without needing to look at memory layout in `gdb` (as well as an implicit explanation of why the offset are as they are in that file).

hybrid/compartment_examples/comp_setup/shared.S
0152la: I'm thinking that might complicate things, and it's fine to have a bit of hard labour here? Just to clarify, you are suggesting deriving these offsets from the structure of `struct comp`, rather than having hardcoded values, is that right?

ltratt: I think at some point we'll change the `struct`, forget to change this `#define`, and then have no idea why things crash in weird ways :)

If we can use `sizeof` from assembly, that would be easiest, but my guess is it's not possible... Maybe there's some other cunning way of keeping the two in sync -- or, at least, putting the size `#define` near the struct so we're more likely to remember to change it?

0152la:
Then how about a question of "what bad things can happen if we do leak this compartment information"? On principle, we don't want to leak it, I agree, but to make our life easier, before we come up with a potentially better solution, can bad things happen if we do?

ltratt:
If the attacker has access to a compartments `ddc` capability (which we currently leak AFAICS), it implicitly can do whatever it wants to that compartment.

hybrid/compartment_examples/comp_setup/shared.S
0152la: I am assuming here we can't, since we're not compiling C code, but just `asm` (@jacobbramley can intervene if I'm incorrect). And when I had the `#define`s in C code, the compiler complained, presumably as they are compiled individually. Also, since these are macros, won't that mean they will just be replaced in `asm` code with what they are defined to? I have yet to try, but I can play around a bit and see. What we might need is have inline `asm` alongside C code, which I think will be a bother...

ltratt: Indeed, it might not be possible! Let's see if @jacobbramley has any ideas; if not, we should give up and move on.

jacobbramley: Neither the assembler nor the C pre-processor understand C struct layouts, so there's no easy way (that I know of). I'm sure something _clever_ is possible, but probably not ideal for an example (unless it's also very clear).

The most robust method that I'm aware of is to `#define` it to a literal (like here) and `(static_)assert` that it matches `sizeof` in a C file somewhere. Likewise with `offsetof` for the fields.

ltratt: > The most robust method that I'm aware of is to #define it to a literal (like here) and (static_)assert that it matches sizeof in a C file somewhere. Likewise with offsetof for the fields.

Aha, that's clever. So that would be in a `.h` file that can be `#include`d in the `.S` file?

0152la: Something like what I did in https://github.com/capablevms/cheri-examples/pull/52/commits/885b2af1214741666e428d41b99b2a21fc9ddc0c ?

ltratt: Yes, I think that's what Jacob's suggesting, although these might as well be `static_assert`s assuming we support C11?

0152la:
> Sp "priviliged".

Sorry, what does this mean?

hybrid/compartment_examples/comp_setup/shared.S
0152la: Updated in https://github.com/capablevms/cheri-examples/pull/52/commits/4c4a6112c827294f8eebfe812c087a46bf98e379, with the "it works on my setup" tag, so I assume it should work otherwise.

ltratt:
> Sorry, what does this mean?

Spelling error: should be "privileged".

hybrid/compartment_examples/comp_setup/main.c
```diff
@@ -113,14 +113,23 @@ add_comp(uint8_t* _start_addr, void (* _comp_fn)())
...
+
 void
 validate_offsets()
 {
-    assert(COMP_SIZE == sizeof(struct comp));
-    assert(COMP_OFFSET_STK_ADDR == offsetof(struct comp, stack_addr));
-    assert(COMP_OFFSET_STK_LEN == offsetof(struct comp, stack_len));
-    assert(COMP_OFFSET_DDC == offsetof(struct comp, ddc));
-    assert(COMP_OFFSET_PCC == offsetof(struct comp, comp_fn));
+    static_assert(COMP_SIZE == sizeof(struct comp),
```
ltratt, line 123: I wonder: do `static_assert`s need to be in a function? If not, we can place these at the top-level of the file when it'll be clearer that they're static thingies.

0152la:
> If the attacker has access to a compartments `ddc` capability (which we currently leak AFAICS), it implicitly can do whatever it wants to that compartment.

I've been mulling over this statement and I think I disagree. In the current setup, the executable bit of a compartment lies in a memory space covered by the PCC, while the memory associated with that compartment lies in a space covered by the DDC. Currently, these are disjoint. Therefore, leaking the DDC would allow an attacker to modify the memory a process reads from/writes to, but it can't modify the execution of that compartment. Although one possible issue would be installing the compartment PCC into the DDC, then it could overwrite functionality. So I don't disagree it's a bad thing to leak the DDC, but rather that leaking "just" the DDC shouldn't give it permission to do "whatever" to the compartment.

Now, there is a discussion to be made if the executable part of a compartment should lie outside the memory that is provided to it. It might be worth raising this issue during today's sync maybe, unless I've misunderstood something about how the PCC/DDC should interact.

ltratt:
Imagine somewhere in your code you have something like:

```
if (somewhere_in_memory->some_boolean) {
  // Do something very dangerous
}
```

if I can flip the `some_boolean` bit in another compartment, I can get it do something very dangerous, in a sense undermining heap compartmentalisation completely.

[A variant on this is that if there are function pointers in another compartment's heap, then I can change them to point to arbitrary gadgets, even in that other compartment's PCC. This will be, in general, harder to exploit than the boolean example though IMHO.]

0152la:
> Sp "priviliged".

I just noticed this was an in-line review on an older commit so it didn't show up in the view I was looking at :clown_face: 

Fixed in https://github.com/capablevms/cheri-examples/pull/52/commits/bb5ac6b0be4e4b6a69b87aaa80a304d988fe6a9b.

hybrid/compartment_examples/comp_setup/main.c
0152la: Updated in https://github.com/capablevms/cheri-examples/pull/52/commits/b93d6d729375eb268f85b93ba49f7d26bb88dbc3.

0152la:
> Imagine somewhere in your code you have something like:
> 
> ```
> if (somewhere_in_memory->some_boolean) {
>   // Do something very dangerous
> }
> ```
> 
> if I can flip the `some_boolean` bit in another compartment, I can get it do something very dangerous, in a sense undermining heap compartmentalisation completely.
> 
> [A variant on this is that if there are function pointers in another compartment's heap, then I can change them to point to arbitrary gadgets, even in that other compartment's PCC. This will be, in general, harder to exploit than the boolean example though IMHO.]

Still not entirely convinced, but this is a conversation we might need to have with some pen/paper/digital draw space.

ltratt:
Let me mildly expand my example, and then I'm going to leave you to chew on it and tell me where I'm wrong in detail :)

My starting assumption is that the DDC must be a capability with memory read/write privileges and bounds for some memory somewhere. Let's then assume I'm executing code in compartment 0. I now can do this:

```
void __capability *compartmentone_heap = compartents[1].ddc;
compartmentone_heap[...] = true;
```

where `...` is the right offset to flip the bit I want in the other compartment. When execution now flips to that other compartment, I have (in a sense) "corrupted" its heap in my favour.

Put another way: if I have access to the DDC then, AFAICS, it's like allowing any Unix process to read/write to any other Unix process's memory.

It's possible I have an incorrect assumption somewhere, however!

0152la:
I agree that it would corrupt the process' memory, and I further agree that is a bad thing. But what I disagree with is "it implicitly can do whatever it wants to that compartment", and "I can get it do something very dangerous". The compartment, with a corrupted memory cannot do more things than it could originally do [1]. So you would need to corrupt the heap, and either (a) be aware of a vulnerability which can be exercised based on specific loaded data, or (b) somehow build around the affected compartment's intended functionality towards some end (which still shouldn't be able to get you higher capabilities, as the attacker is still constrained within its own bounds).

[1] actually, it probably could if it,for some forsaken reason, stores executable code within its memory, which it loads, and then executes, if that is even possible first of all, and if the PCC doesn't mind such things somehow.

ltratt:
>  The compartment, with a corrupted memory cannot do more things than it could originally do [1]

Let's take this as our definition. Let's imagine I have two booleans that interact, and there are 3 valid states that my code copes with. If another compartment comes in and mutates those booleans to force it to the 4th state that my code doesn't cope with, my compartment is in a real sense doing something it wasn't designed to do.

0152la:
Correct, if we abstract the compartment as a state machine. That could potentially fall under case (a) that I described above. Ok, I shall concede that modifying the DDC can affect reachable states, which can cascade to putting the end-state of a compartment execution into something not "valid".

However, I shall be very pedantic and still say that the compartment contains that execution even before memory corruption. Whether or not inputs are considered "valid" means nothing if they are not validated. As such, I would argue that this is an instance of exercising a vulnerability, rather than expanding the functionality. Nevertheless, it is very much correct that not leaking the DDC would prevent this vulnerability from being exercised. Therefore, I can, on my end, conclude that I am content with this result.

ltratt:
> As such, I would argue that this is an instance of exercising a vulnerability, rather than expanding the functionality. Nevertheless, it is very much correct that not leaking the DDC would prevent this vulnerability from being exercised.

I am in agreement with both points.

Do we want to tackle this in this PR or do that in a follow-up? I'm fine either way.

0152la:
I think strengthening security should be done after the functionality of compartment changing has been polished properly, so let's leave it for a few PRs from now I would say. We already had some discussion way beyond the initial bounds of the PR.

ltratt:
I think we probably want to tackle it soon, but I'm fine with not doing it in this PR.

Please squash.

0152la:
Before squash, a quick `clang-format` @ltratt .

ltratt:
Please squash.

0152la:
Squashed.

PR #53, opened 2021-12-03
Add example showing inter-compartment switch

0152la:
This example adds a second compartment, and shows a possible design to
achieve inter-compartment switching without (much) privilege escalation.
In this file, we setup compartment information within array `comps`
(`main.c:84`). This information will be used by the assembly function
`switch_compartment`. Further, we save a capability to
`switch_compartment()` in a register (`main.c:102`), to be able to call
it from a bounded-PCC context, as well as a capability defining the area
of memory with compartment information (`main.c:101`). These could be
added to the memory area of each created compartment for better
security, as well as sealed.

Our two compartments provide one entry point each, `comp_f_fn`
(`shared.S:93`), and `comp_g_fn` (`shared.S:109`). The design is that
compartment `f` will perform a switch to compartment `g`, which performs
some observable action (in this instance, sets a specific memory
location to a specific integer value). The switching process in (almost)
entirely bound within `switch_compartment`, on the execution side, and
the memory area with compartment information, for memory access. We note
almost, as the executable area is approximated (`main.c:96`).

### DIFF
--- a/hybrid/compartment_examples/inter_comp_call/main.c
+++ b/hybrid/compartment_examples/inter_comp_call/main.c
@@ -1,0 +1,152 @@
+/* This example adds a second compartment, and shows a possible design to
+ * achieve inter-compartment switching without (much) privilege escalation. In
+ * this file, we setup compartment information within array `comps`
+ * (`main.c:84`). This information will be used by the assembly function
+ * `switch_compartment`. Further, we save a capability to
+ * `switch_compartment()` in a register (`main.c:102`), to be able to call it
+ * from a bounded-PCC context, as well as a capability defining the area of
+ * memory with compartment information (`main.c:101`). These could be added to
+ * the memory area of each created compartment for better security, as well as
+ * sealed.
+ *
+ * Our two compartments provide one entry point each, `comp_f_fn`
+ * (`shared.S:93`), and `comp_g_fn` (`shared.S:109`). The design is that
+ * compartment `f` will perform a switch to compartment `g`, which performs
+ * some observable action (in this instance, sets a specific memory location to
+ * a specific integer value). The switching process in (almost) entirely bound
+ * within `switch_compartment`, on the execution side, and the memory area with
+ * compartment information, for memory access. We note almost, as the
+ * executable area is approximated (`main.c:96`).
+ */
+
+#include "../../../include/common.h"
+#include "../../include/utils.h"
+
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(__CHERI_CAPABILITY_WIDTH__) || defined(__CHERI_PURE_CAPABILITY__)
+#error "This example only works on CHERI hybrid mode"
+#endif
+
+/*******************************************************************************
+ * Assembly Functions & Consts
+ ******************************************************************************/
+
+extern int switch_compartment();
+extern void comp_f_fn();
+extern void comp_g_fn();
+
+/*******************************************************************************
+ * Types & Consts
+ ******************************************************************************/
+
+const size_t comp_stack_size = 2000;
+const size_t total_comp_size = 5000;
+size_t id = 0;
+
+/* Abstract representation of a compartment. Within 80 bytes we represent:
+ * - an id (8B)
+ * - address to the start of the compartment's stack (8B)
+ *   the size of the stack (8B)
+ * - address to the start of the compartment's heap (8B)
+ *   the size of the heap (8B)
+ * - alignment padding (8B)
+ * - the ddc corresponding to the compartment (16B)
+ * - the function within the compartment to be executed upon switching (16B)
+ */
+struct comp
+{
+	size_t id;
+	void *stack_addr;
+	size_t stack_len;
+	void *heap_addr;
+	size_t heap_len;
+	void *__capability ddc;
+	void *__capability comp_fn;
+};
+
+// ASM offsets, included here for validation
+#include "main.h"
+
+static_assert(COMP_SIZE == sizeof(struct comp), "Invalid `COMP_SIZE` provided");
+static_assert(COMP_OFFSET_STK_ADDR == offsetof(struct comp, stack_addr),
+			  "Invalid `COMP_OFFSET_STK_ADDR` provided.");
+static_assert(COMP_OFFSET_STK_LEN == offsetof(struct comp, stack_len),
+			  "Invalid `COMP_OFFSET_STK_LEN` provided.");
+static_assert(COMP_OFFSET_DDC == offsetof(struct comp, ddc), "Invalid `COMP_OFFSET_DDC` provided.");
+static_assert(COMP_OFFSET_PCC == offsetof(struct comp, comp_fn),
+			  "Invalid `COMP_OFFSET_PCC` provided.");
+
+struct comp comps[COMP_COUNT];
+
+/*******************************************************************************
+ * Privileged Functions
+ ******************************************************************************/
+
+/* This function stores the compartment data in memory, so that it can be
+ * accessed by the compartments as needed.
+ */
+void executive_switch(struct comp c)
+{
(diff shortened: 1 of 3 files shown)


hybrid/compartment_examples/inter_comp_call/main.c
```diff
@@ -0,0 +1,151 @@
...
+ */
+void executive_switch(struct comp c)
+{
+    void* __capability switch_cap = (void* __capability) switch_compartment;
+    switch_cap = cheri_bounds_set(switch_cap, 80 * 4);
+
+	void* __capability comps_addr = (void* __capability) &comps;
+    comps_addr = cheri_bounds_set(comps_addr, COMP_COUNT * COMP_SIZE);
+
+	asm("mov c19, %w0\n\t"
```
ltratt, line 101: I think there might be a mix of tabs and spaces here? Let's use all spaces.

0152la: Ah, well, this will be fixed by the subsequent `clang-format` process (which I haven't applied for the PR at this point). I prefer spaces myself, but `clang-format` seems to put tabs.

ltratt:
You're going to have to spell out a few things in more details so that idiots like me can understand what's going on :) For example, what does "shows a possible design to achieve inter-compartment switching without (much) privilege escalation" mean? I guess this PR is still subject to the same "any compartment can get access to another compartments DDC and read/write to its heap" as the previous PR?

hybrid/compartment_examples/inter_comp_call/main.c
ltratt: Does it? Ye gods! In general, it's best to do `clang-format` before opening a PR. And then if it does mix tabs and spaces we'll just have to live with it (we can still grumble about it though :)).

0152la: For ease, I've just ran one now [1]. In the future, I will run one at the point of making the PR, and the point of merging (until I hopefully am able to do the hooks thing).

[1] https://github.com/capablevms/cheri-examples/pull/53/commits/5ae74cf0bc9ad355c7ff59bdc22b328bd242065a

ltratt: Thanks!

0152la:
> I guess this PR is still subject to the same "any compartment can get access to another compartments DDC and read/write to its heap" as the previous PR?

Yes, however I briefly mention [1]:

> These could be added to the memory area of each created compartment for better security, as well as sealed.

This might be a good direction for a future PR, or even this one, to make it slightly meatier. 

> You're going to have to spell out a few things in more details so that idiots like me can understand what's going on

The top-level comment is a high-level view, so it's kept intentionally less exact (to not be overly verbose with detail), while the steps are more spelled out in the in-line comments. Do those not help?

[1] https://github.com/capablevms/cheri-examples/pull/53/files#diff-bd3f0e982085d3f24aabff1a3429bddce60c0605c9d6047f8354c515a685b477R8

0152la:
I forgot to mention in the leading comment, there are a things I'd like to improve. Primarily, the PCC bounds on `switch_compartment` are approximate and manual [1], as well as the bounds on the compartment functions [2]. However, I am unaware of how to derive these in code (C or asm), without doing something similar as ELF inspection that the Android example does.

And the security can  (and should) be tightened further, primarily via sealing capabilities for the compartment switcher (DDC/PCC) and copying them in memory for each specific compartment. I have an idea of how to do that as well in here, and could integrate it in this PR, or a separate "security strengthening" pass.

[1] https://github.com/capablevms/cheri-examples/pull/53/files#diff-bd3f0e982085d3f24aabff1a3429bddce60c0605c9d6047f8354c515a685b477R96
[2] https://github.com/capablevms/cheri-examples/pull/53/files#diff-bd3f0e982085d3f24aabff1a3429bddce60c0605c9d6047f8354c515a685b477R128

ltratt:
>  These could be added to the memory area of each created compartment for better security, as well as sealed.
>
> This might be a good direction for a future PR, or even this one, to make it slightly meatier.

My guess is that the easiest way of handling this is for the compartment switcher to keep these things private to itself. But we can handle that in a later PR.

> I have an idea of how to do that as well in here, and could integrate it in this PR, or a separate "security strengthening" pass.

Sounds like a good idea for the next PR :)

hybrid/compartment_examples/inter_comp_call/main.c
```diff
@@ -0,0 +1,150 @@
...
+ * Privileged Functions
+ ******************************************************************************/
+
+/* This function stores the compartment data in memory, so that it can be
+ * accessed by the compartments as needed.
+ */
+void executive_switch(struct comp c)
+{
+	void *__capability switch_cap = (void *__capability) switch_compartment;
+	switch_cap = cheri_bounds_set(switch_cap, 80 * 4);
```
ltratt, line 96: Ah, are these used as PCC bounds?

hybrid/compartment_examples/inter_comp_call/main.c
```diff
@@ -0,0 +1,150 @@
...
+	new_comp.stack_len = comp_stack_size;
+	new_comp.heap_addr = (void *) (_start_addr + comp_stack_size);
+	new_comp.heap_len = total_comp_size - comp_stack_size;
+
+	void *__capability comp_ddc = (void *__capability) _start_addr;
+	comp_ddc = cheri_bounds_set(comp_ddc, total_comp_size);
+	new_comp.ddc = comp_ddc;
+
+	void *__capability comp_fn = (void *__capability) _comp_fn;
+	comp_fn = cheri_bounds_set(comp_fn, 40);
```
ltratt, line 128: I guess these are also PCC bounds?

hybrid/compartment_examples/inter_comp_call/main.c
0152la: Yes, hardcoded approximate bounds so things can execute, since I'm unaware of a way to compute this at compile-time, if it is possible. They would potentially be handled by the linker, an an actual implementation. Could maybe use labels, to mark start/end?

hybrid/compartment_examples/inter_comp_call/main.c
0152la: Yes, bounds as well.

hybrid/compartment_examples/inter_comp_call/main.c
ltratt: Good question -- I have no idea! Certainly it should be possible via ELF but it might be possible some other way, though I doubt that it's portable either way. I'm fine with punting on that for this PR. Maybe add an issue to this repo "How to calculate a function's length at runtime" as we'll then be able to ask Edd for suggestions (he knows more about such things than I do!).

hybrid/compartment_examples/inter_comp_call/main.c
ltratt: Can I suggest we use `pcc` somewhere in the var name? Maybe something like `comp_pcc` or similar? Failing that, a comment will help readers like me.

hybrid/compartment_examples/inter_comp_call/main.c
0152la: Opened #54 to at least start a conversation maybe.

hybrid/compartment_examples/inter_comp_call/main.c
0152la: I've added a comment which helpfully clarifies things.

At this point, I think the following is a rather core property of CHERI, and maybe we should make it obvious somewhere: creating a capability which has the `value` field the address of a function, then calling (or jumping to) that capability implies installing that capability as the PCC.

hybrid/compartment_examples/inter_comp_call/main.c
ltratt: Thanks!

ltratt:
Please squash.

0152la:
Squashed.

Issue #54, opened 2021-12-06
Calculating size of function


0152la:
Suppose we have a function definition, in either C or assembly. Is it possible to calculate the size of such a function (presumably in byes) at compile time? One potential solution would be to have the functions whose sizes we want to calculate in a separate file, compile it, and perform the computation based on ELF data, but are there maybe easier solutions?

jacobbramley:
A typical method (e.g. used by `clang` and others) is to define a (local) symbol at the end of the function and set [`.size`](https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_7.html#SEC123) based on arithmetic. For example:

```
    .globl  func
    .p2align        2
    .type   func,@function
func:
    ...
    ret
.Lfunc_end0:
    .size    func, .Lfunc_end0-func
```

It might not be feasible to read that from source — I've never tried — but for assembly functions you can define a global symbol (instead of the local `.Lfunc_end0`) or perhaps calculate the size and store it as some other global constant.

0152la:
An approach of doing this was implemented in #55. Do we still want to keep this issue for potential future improvements to the current implementation?

jacobbramley:
In principle we could do better — making it a static constant with a relocation — but whether or not that is achievable is unclear. It's probably a low priority concern, anyway.

PR #55, opened 2021-12-16
Improve compartment example security

0152la:
In this example, we tighten security even further. For each compartment we
create, we store some capabilities of interest. Essentially, we setup
certain entry points, and make them available to each compartment.
Currently, these involve the PCC/DDC of the `switch_compartment` function
(PCC to execute, DDC to access memory where we store information pertaining
to compartments; essentially, `switch_compartment` performs a limited
privelege escalation, but only over prepared data within `main`).
Additionally, we also seal the PCC of `switch_compartment`.


0152la:
I'm more or less putting this here now to see if anyone has any ideas if we can break this model. I'll probably take a bit of time to tinker with that. I also have not yet implemented exact function bounds (as per #54), so that is one know vector of attack. But as is, I believe a compartment can now not access data regarding other compartments (which is available within `switch_compartment`).

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
+/* In this example, we tighten security even further. For each compartment we
+ * create, we store some capabilities of interest. Essentially, we setup
+ * certain entry points, and make them available to each compartment.
+ * Currently, these involve the PCC/DDC of the `switch_compartment` function
+ * (PCC to execute, DDC to access memory where we store information pertaining
+ * to compartments; essentially, `switch_compartment` performs a limited
+ * privelege escalation, but only over prepared data within `main`).
```
ltratt, line 7: sp "privilege"

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+ * Additionally, we also changed the expected memory layout in the
+ * compartments. In the absence of a custom memory allocator, we note that the
+ * heap is not used, except for manually storing capabilities as detailed
+ * above. Current compartment memory layout:
+ *
+ * `DDC value` > -----------------
+ *              |                 |
+ *              |     HEAP        |
+ *              |                 |
+ *              |   saved caps    |
```
ltratt, line 26: What is included in "saved caps"? I guess these are the "some capabilities of interest" mentioned earlier, but I must admit that I don't know precisely what that set might be!

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+/*******************************************************************************
+ * Types & Consts
+ ******************************************************************************/
+
+const size_t comp_stack_size = 1992;
+const size_t total_comp_size = 5000;
+size_t id = 0;
+
+// Duplicate certain capabilities within the heap of each compartments;
+// currently we save DDC and PCC of `switch_compartment`
```
ltratt, line 63: If another compartment can arbitrarily read/write to the compartment switcher I think, in general, that it can get the compartment switcher to do whatever it wants. But, if `switch_compartment` is a sentry (which I *think* it is), we don't need its PCC to be stored outside? The DDC is an interesting question: I'm not sure how to handle that. Can a sentry also change the DDC?

hybrid/compartment_examples/inter_comp_call_secure/shared.S
```diff
@@ -0,0 +1,171 @@
...
+
+    // Load DDC
+    add       x11, x10, #COMP_OFFSET_DDC
+    ldr       c1, [x19, x11]
+
+    // Setup SP
+    mov       x12, sp
+    add       x11, x10, #COMP_OFFSET_STK_ADDR
+    ldr       x11, [x19, x11]
+//    add       x13, x10, #COMP_OFFSET_STK_LEN
```
ltratt, line 44: In general we prefer to delete commented code (unless there's a really good reason to keep it around).

0152la: Whoops. Removed in https://github.com/capablevms/cheri-examples/pull/55/commits/3955d6af194c6f59f5bc9f71b06702e66ecaa9dc.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Fixed in https://github.com/capablevms/cheri-examples/pull/55/commits/3955d6af194c6f59f5bc9f71b06702e66ecaa9dc.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: This is detailed in `main.c:4`:
> Currently, these involve the PCC/DDC of the `switch_compartment` function (PCC to execute, DDC to access memory where we store information pertaining to compartments

Further, there is a `saved_caps` variable a bit down. I can add in an underscore to make the connection more obvious?

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Sorry, I don't quite follow. I think there are three issues at play here:
1. The minimal amount of data a malicious compartment requires in order to be able to "take control" of `switch_compartment` (sentence 1).
2. Requiring the PCC of `switch_compartment` to be "stored outside" (I don't understand where "outside" is). (sentence 2)
3. Issue of DDC integrity (sentence 3-4)

For 1, we need to balance the minimal information `switch_compartment` needs in order to actually perform the switch. We obviously need *something* to be exposed in order for the function to be executable (at minimum, we can say the address of `switch_compartment`, but of course that is useless without knowing what compartment we want to switch to, or where that compartment is in memory). If the minimal information needed to perform the switch and the minimal data needed to take control of `switch_compartment` are not disjoint, then this is a design problem, and this design is infeasible.

For 2, I will clarify that the `switch_compartment` available to the compartments is a sentry, which can only be branched to. I need further clarification regarding the "outside" thing. For 3, what does "a sentry [..] change the DDC" mean? A sentry is a sealed capability with it's value being the address of a function. By that definition, a sentry is changing the DDC, as the DDC swapping happens inside `switch_compartment` (although the DDC is expected in a register, provided as an argument by the calling compartment).

0152la:
After further talks, I believe we conclude that leaking the DDC of the compartment switcher is unacceptable. Therefore this design is inappropriate in its current incarnation, unless we can somehow communicate compartment information to the switcher function without needing to expose this DDC capability.

ltratt:
I think that's a fair summary of our current thoughts.

0152la:
Would you say sealing the DDC (to be unsealed somehow in `switch_compartment` work?

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+
+/* This function stores the compartment data in memory, so that it can be
+ * accessed by the compartments as needed.
+ */
+void executive_switch(struct comp c)
+{
+	void *__capability comps_addr = saved_caps[1];
+
+	asm("mov c19, %w0\n\t"
+		"mov x0, #0\n\t"
```
jacobbramley, line 126: You need to make sure (basically) that the C compiler doesn't mess with those registers between setting them, and reading them in `switch_compartment`. The only safe way to do that is to make sure that there is no C code in between, which mean writing this whole function in assembly.

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+		"msr CID_EL0, c0"
```
jacobbramley, line 127: More generally — and this requires some thought — does splitting the actual switching not represent a security concern? For example, what happens if I call `switch_compartment` directly? Nothing seems to prevent me from doing so, and then I can provide whatever value in `c19` that I want.

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+	struct comp new_comp;
+	new_comp.id = id;
+
+	new_comp.stack_addr = (void *) (_start_addr + total_comp_size - comp_stack_size);
+	new_comp.stack_len = comp_stack_size;
+	new_comp.heap_addr = (void *) _start_addr;
+	new_comp.heap_len = total_comp_size - comp_stack_size;
+
+	// Memory magic; we will store capabilities at the memory location pointed
+	// by `heap_addr` of current compartment
```
jacobbramley, line 158: Wouldn't that corrupt the heap?

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+
+	void *__capability comps_addr = (void *__capability) &comps;
+	comps_addr = cheri_bounds_set(comps_addr, COMP_COUNT * COMP_SIZE);
+	saved_caps[1] = comps_addr;
+}
+
+/* This function stores the compartment data in memory, so that it can be
+ * accessed by the compartments as needed.
+ */
+void executive_switch(struct comp c)
```
jacobbramley, line 121: I can't see that this actually uses executive mode. What have I missed?

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley, line 127: I suspect that passing this in a general-purpose register might be better. It can be set in `switch_compartment`, rather than set here and read there.

The real answer here rather depends on what the ID actually means.

jacobbramley (commented):
I think I'll need another pass for my own understanding. I'm not sure if I yet understand what guarantees each compartment (including the switcher) can reasonably make, for example.

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley: > although the DDC is expected in a register, provided as an argument by the calling compartment

I think that's a concern. What stops the caller from providing an inappropriate DDC? However, in this case I think it can only cause a crash in the switcher (by providing an inadequate DDC). That might be acceptable in the (currently implicit) threat model.

If it's necessary to bind a pair (to prevent this), it's possible to do that using `ldpblr`, as we've discussed before.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Yes, that is a concern. The problem with the current design is that in order to access the compartment information (which we want to keep secure from compartments) in memory, we need to set the DDC over that region of memory. I can't think of a way to self-determine this in `switch_compartment`, without some expected cooperation from the compartment we are transitioning from.

As for why you can't call `switch_compartment` directly (assuming from within a compartment), it is that the PCC for the compartment you are in does not cover `switch_compartment`, and the only way to transition is via the provided sentry.

0152la: Yes, honestly, `c19` is more a remnant of previous iterations. It can be anything, or if we can change the design of `switch_compartment` to not require this argument be passed somehow...

If by ID you mean the contents of `CID_EL0`, it represents the ID of the compartment we are jumping into (which would, of course, require some knowledge between compartments of what compartment corresponds to what ID). This currently corresponds to the index of the `comps` array.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: This "executive" is not related to the notion of executive/restricted in Morello, just bad naming we could say. I will edit it to make sure there is no such confusion.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: I'm unsure what corrupt the heap implies, but it is the case that we consider the heap region where we store this info as "reserved", and ideally a memory allocator would avoid overwriting this.

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+struct comp comps[COMP_COUNT];
+
+/*******************************************************************************
+ * Privileged Functions
+ ******************************************************************************/
+
+void init_comps()
+{
+	void *__capability switch_cap = (void *__capability) switch_compartment;
+	switch_cap = cheri_bounds_set(switch_cap, 80 * 4);
```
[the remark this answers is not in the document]
0152la: Yes, this is related to #54 . I have not implemented this yet, but at least lifting it for the future would be good.

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley: I meant `CID_EL0`. One of the first things that `switch_compartment` does is to read the value out of `CID_EL0` (into `c10`), to use it to index the `comps` array, but then why not just leave it in a general-purpose register (`c0`) and let `switch_compartment` choose if or how to put it in `CID_EL0`? That seems like a better abstraction to me, and a clearer interface because it looks more like a conventional function call.

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley: I think that's my point: if the allocator avoids overwriting it, then it's not really part of the heap, and instead is a region in its own right.
Note that this won't support re-entrant or multi-threaded compartments that might need to have multiple calls active at any one time, but that's Ok here as long as the limitation is noted.

0152la:
I have pushed a new commit making use of `ldpblr`, as discussed, and obfuscating the switcher's required capabilities.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Since the only thing following is the call to `switch_compartment`, would it be safe to simply include that function call as part of the inline assembly, maybe?

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: That seems to be a problem to consider when we implement memory management for these compartments, in however fashion that might be. For this example, I believe what we call "compartment heap" has no actual meaning, so, if we do mention it (as I believe I have in the intro comment), it should be fine?

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Indeed, there is no practical reason, beyond showing one way of using this "special-purpose-but-not-really" CID register. I can either add a comment to emphasise this, or we can use `c0` or any other register, if we prefer to have a "cleaner" example.

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley: I'm arguing only for calling it something else, since this register storage area is fundamental to the switching mechanism. The actual behaviour is fine, as you point out, but this isn't a "heap_addr" (yet).

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley: You can, but you'll have to mark all potentially-clobbered registers as such, and there are a lot of them. (All the caller-saved registers, basically.) It can be done if that's what you need to do, but it's often easier to put the whole lot into a `.s` (or `.S`) file.

hybrid/compartment_examples/inter_comp_call_secure/main.c
jacobbramley: At the time of writing, I would have argued that holding the sentry is sufficient to make the call, since a malicious caller could provide a mis-matched DDC (or ID from which it is derived, or similar). However, after our recent call I now think you've resolved that using `ldpblr` anyway. Is that the case?

jacobbramley: I think it's good to set it, but we should do so in `switch_compartment`. Here, we should just pass the value in `c0` (or similar). I don't think using `CID_EL0` as a parameter-passing register would be considered a good idea.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: I see. So the naive solution then is to move `executive_switch` in our `shared.S` file, or is that still insufficient?

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: That is correct - the compartment would have the PCC (albeit sealed), so it could call `switch_compartment` and give it a malformed value in `c19` (now `c29`), which could lead to something unwanted. Now, the compartment only is aware of a sealed capability pointing to some address (which happens to contain the unsealed switcher DDC), but that capability can only be used via `lpb`-like instructions (including `ldpblr`), with set parameters (i.e., the DDC/PCC combo set in memory, outside of the compartment's DDC bounds).

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Ah, I think I missed to explain a point. It just so happens that the memory location I want to place this capability is contained within the `heap_addr` variable, which determines where the compartment's highest heap address is. That's why I'm using this variable, named `heap_addr` in the `struct`.

Actually I just realised I made a mistake in the memory design of the compartments. In the diagram [1], I have lower addresses at the top, when I meant to have higher ones. The idea is to store this capability at the highest heap address, far away from where user-affected memory would lie.

[1] https://github.com/capablevms/cheri-examples/pull/55/files#diff-1ba17d80ea42080e89335c8f51d44a66e80fe3df28b74490c2f8e4a36ac1c786R28

ltratt:
This looks very promising! Can we create some "negative examples" (for want of a better term) in `hybrid/compartment_examples` that show what happens when code tries doing naughty things? For example, we could have a few examples which try to read the DDC from `switch_compartment` and check that they fail.

[At the moment we lack a nice testing infrastructure on CHERI for this stuff. @drpdr used shell scripts elsewhere in the repo to automatically check which programs should succeed and which should fail -- that's still the best approach I can think of.]

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
+/* In this example, we tighten security even further. For each compartment we
```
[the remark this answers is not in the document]
0152la: Clarified in https://github.com/capablevms/cheri-examples/pull/55/commits/9fe7172cb480ec112d21e24f84656c559fee9f3b.

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -0,0 +1,194 @@
...
+size_t id = 0;
+
+// Duplicate certain capabilities within the heap of each compartments;
+// currently we save DDC and PCC of `switch_compartment`
+const size_t saved_caps_count = 2;
+void *__capability saved_caps[saved_caps_count];
+
+/* Abstract representation of a compartment. Within 80 bytes we represent:
+ * - an id (8B)
+ * - address to the start of the compartment's stack (8B)
```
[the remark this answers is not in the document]
0152la: Clarified in https://github.com/capablevms/cheri-examples/pull/55/commits/db7617bdb8a0bfc3b6cec9413d285b91f5b2a000.

0152la:
Another rather meaty update, fixing memory allocation within the compartment (again, for no current functional use), and a few other tidbits. I think the only thing left is to remove the usage of `CID_EL0`; I'll want to do so probably early tomorrow, along with a `clang-format`, to use the new version of the build in the CI, as per our earlier remarks (I have not updated the build as I was working on the PR). Otherwise, if I missed anything, it would be useful to resolve some of the conversations we have lying around, since it's really hard to properly look for them with all these outdated versions of the PR.

hybrid/compartment_examples/inter_comp_call_secure/main.c
```diff
@@ -146,22 +151,25 @@ void executive_switch(struct comp c)
...
 
+    // Ensure 16-byte alignment throught the compartment bounds
+    assert(((uintptr_t) new_comp.compartment_start) % 16 == 0);
+    assert(((uintptr_t) new_comp.stack_addr) % 16 == 0);
+    assert(total_comp_size % 16 == 0);
+
     // When creating a compartment, store a local copy of the capability which
     // will allow us to call `switch_compartment` in the heap of the compartment.
-    memcpy(new_comp.heap_addr, &switcher_call, sizeof(void *__capability));
+    void * heap_top = (void *) (_start_addr + total_comp_size - sizeof(void* __capability));
```
ltratt, line 173: Minor formatting: s/`void * heap_top`/`void *heap_top`/ and s/`void* __capability`/`void *__capability`/.

0152la: These should be fixed once I apply `clang-format`, which I have not quite yet. Should be shortly.

0152la:
I think that we should finalize this PR first, considering both its age and the amount of discussion around it, before adding negative examples in a follow-up PR. However, I have nothing against including those here, if we think that's reasonable.

ltratt:
Let's add 1 simple negative example to this PR, just as a proof of concept, then I agree: we can merge, and the next PR can add lots more negative examples.

hybrid/compartment_examples/inter_comp_call_secure/main.c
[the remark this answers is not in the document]
0152la: A better (not necessarily good) way of getting function sizes has been added in https://github.com/capablevms/cheri-examples/pull/55/commits/db7617bdb8a0bfc3b6cec9413d285b91f5b2a000.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Removed the use of `CID_EL0` as discussed in https://github.com/capablevms/cheri-examples/pull/55/commits/47e39b76d2af704cb87d2bf7653a3b2c5b032d47.

hybrid/compartment_examples/inter_comp_call_secure/main.c
0152la: Applied `clang-format` in https://github.com/capablevms/cheri-examples/pull/55/commits/5d529edf16f81593d7d5c93db5ab868e4937f978.

0152la:
Added a negative example, where we try to dereference a local capability for `switch_compartment` access. The dereference predictably fails due to it being sealed.

I believe at this point, the only unresolved issue [1] is moving `executive_switch` from `C` to the assembly file, in order to ensure no clobbering of the registers we need set up. I'll have a go at that.

[1] https://github.com/capablevms/cheri-examples/pull/55#discussion_r771482397

ltratt:
Please squash as you see fit.

0152la:
I've also pushed the final change of moving `executive_switch` in assembly, to prevent potential register clobbering at the level of C. I'll squash after the meeting today, just in case there are any further things left to do.

Furthermore, I've tried running both this example, as well as some of the previous ones on our daily CHERI build (with @drpdr's help), but I encountered some issues. As of now, none of the examples I have added (after `ddc_compartment_switching_sentry`) have been, or will be, executed outside my local dev environment.

ltratt:
What issues? We should fix them before merging unless there's a very good reason not to.

0152la:
No issues with the infrastructure itself, but difficulties in trying to manually make use of it. The final step was to get the correct SSH key to connect to, which I'll bother Pino about at the some point. Just that, overall, trying to manually use our nightly builds could be more streamlined.

The one "issue" I would say it has is that it runs a very limited set of examples, which need to be manually provided [1]. This includes not ensuring that a newly provided example does indeed run on the build (but it does check that it compiles, as I was told). I would put this with overall infrastructure improvements we could be making to the repo, as we should also improve the overall organization and build system.

[1] https://github.com/capablevms/cheri-examples/blob/master/tests/run_tests.sh#L61

ltratt:
I agree that we can improve things: but we should hook this new example into `buildbot.sh` (directly or indirectly) in this PR. That way we will make sure that we pick up future enhancements to this code too :) If you push a commit to that effect, I think we can get this merged.

ltratt:
The good thing about pushing it is that bors won't merge it if it doesn't work! So go ahead.

0152la:
Good point. I've added this, and `inter_comp_call` to the tests.

ltratt:
Excellent! Since you're not sure if they work or not I recommend you do a `bors try` first to see if it works.

0152la:
bors try

ltratt:
Our standard practise in this situation is to keep pushing fix commits and doing `bors try` until it works :) [You can see the failure by clicking on the "buildbot/capablevms-test-script" test script FWIW.]

0152la:
bors try

ltratt:
@0152la Time to squash?

0152la:
@ltratt I assumed, after our conversation, that I would integrate the failing example in this PR. I can do that when I add the rest of the examples in the CI, in a future PR then, or I can just update this PR tomorrow if we want the failing example in as well.

ltratt:
Ah, sorry, I thought the failing example was in here! Let's link it in -- we might as well :)

0152la:
After trying lots of things, I decided, at the moment, to just have some duplication and move some stuff in their independent directories. I will do some tries now to see if I got everything right.

0152la:
bors try

0152la:
bors try

0152la:
bors try

0152la:
@ltratt @jacobbramley I think we're set. This final set of commits reorganized 3 examples: the main correct example of this PR, the negative example of this PR, and the one in #53. It also adds all these three examples to our CI.

ltratt:
Unless @jacobbramley has any comments, please squash.

0152la:
Squashed; can of course address further comments, if there are any.

hybrid/compartment_examples/inter_comp_call/secure-try_deref/main.c
```diff
@@ -0,0 +1,203 @@
...
+
+	void *__capability comp_ddc = (void *__capability) _start_addr;
+	comp_ddc = cheri_bounds_set(comp_ddc, total_comp_size);
+	new_comp.ddc = comp_ddc;
+
+	// Set up a capability pointing to the function we want to call within the
+	// compartment. This will be loaded as the PCC when the function is called.
+	void *__capability comp_fn = (void *__capability) _comp_fn;
+	// 40 is arbitary; meant to be the size of the executable function within
+	// compartment
```
jacobbramley, line 177: I think you addressed that with precise pointer arithmetic.

jacobbramley (commented):
It looks roughly correct but I'll look more closely later. I think all the earlier threads are resolved.

hybrid/compartment_examples/inter_comp_call/secure-try_deref/main.c
0152la: Ah, I forgot to update this comment. I'm unsure if we can amend this PR as is, after merging, or should this be done as part of another PR? @ltratt

ltratt: This is merged, so it can be done in another PR. I'm fine with small PRs that fix such issues!

PR #56, opened 2022-01-14
Fix `switch_compartment` PCC bounds

0152la:
A hardcoded value of the bounds over `switch_compartment` meant that the
bounds were much wider than expected. This meant two things:
- the `clean` function called within `switch_compartment` was reachable
  and executable;
- when deriving the `clr` within `switch_compartment`, we would derive
  with the larger bound, and return to a compartment with much more
  executable permissions than intended.

We now tightly bind the PCC to `switch_compartment`. This means moving
`clean` within these bounds (a better way would be to create
capabilities allowing this function to be called, and provide local
copies to each compartment, but that is beyond the scope of this
example, and an exercise in engineering), and not deriving `clr` within
`switch_compartment`, but just retaining the provided `clr`.

Also some comment inconsistency fixes.


0152la:
I'm also working on updating the negative example, as of course it's not as easy as copy pasting, and needs to be debugged...

ltratt:
This looks good to me. @jacobbramley are you able to quickly take a peek?

0152la:
Also a way of maybe splitting the assembly code in multiple files, so there is less duplication, and less stuff I have to change when I modify core parts. I'll have to play with `#include`s and makefiles and see.

hybrid/compartment_examples/inter_comp_call/secure/main.c
```diff
@@ -63,6 +63,7 @@ extern void comp_f_fn();
 extern void comp_g_fn();
 extern void *comp_f_fn_end;
 extern void *comp_g_fn_end;
+extern void *switch_compartment_end;
```
jacobbramley, line 66: The types of these `_end` variables aren't really correct. These are the types that they point to, which here is just a zero-sized part of the code (`.text` section). There isn't a good, correct choice, but perhaps representing them as a function pointer is best for this example. Just don't call them; that won't end well. (It shouldn't be possible if the compartment is set up correctly.)

I think this caused your earlier question about taking the address. It's surprising to take the address of a global `void*` but not so much to take the address of a function, especially given the arithmetic where this is used.

hybrid/compartment_examples/inter_comp_call/secure/shared.S
```diff
@@ -53,14 +58,11 @@ switch_compartment:
     ldr       x11, [x29, x11]
     mov       sp, x11
 
-    // Derive a new clr to restore PCC, and store it.
-    cvtp      c11, lr
-
     // Install compartment DDC
     msr       DDC, c1
 
     // Save old DDC (c2), old SP (x12), old LR (c11) on stack
```
jacobbramley, line 64: The comment (`c11`) is now out of date.

0152la: Fixed in https://github.com/capablevms/cheri-examples/pull/56/commits/8ca60bd06adbbe2c55ab81f33a6cd537b175b4c5.

hybrid/compartment_examples/inter_comp_call/secure/main.c
0152la: I was thinking address values are usually represented as `void*`, but if you think functions are better, I have nothing against that. Amended in https://github.com/capablevms/cheri-examples/pull/56/commits/1b46b9cebdf195bcedad7e24ee673a9450701388.

jacobbramley: They aren't addresses, though. If anything, they're `void`s. The label points to the value (which has zero size) and you take the address (`&switch_compartment_end`) to do some pointer arithmetic.

0152la: I'll maybe have to read more on this, but I believe the current incarnation should be sufficient at the time, until we find a "proper" way (if one such way exists). Or maybe improve how we calculate function sizes in #54 .

ltratt: @jacobbramley Are you suggesting using something other than `void *` e.g. `char *` or `u8*`?

jacobbramley: I would declare it as a function ...
```
extern void switch_compartment_end();
```
... because a function has no size as far as C is concerned. All the other things do (except `void`, but I think that'll be rejected). This also works better in C64 because bit zero can be set in _both_ so that the size calculation will still work correctly. It doesn't matter now, but might do if this is adapted later.

This also makes a little more sense in the arithmetic semantics, since the `&` becomes implicit and then you're subtracting one function pointer from another, rather than a function pointer from a `void**`:

```
//                                 v-- '&' elided here because it's implicit for functions.
size_t switcher_size = `(uintptr_t) switch_compartment_end - (uintptr_t) switch_compartment`;
```

All of these options are _undefined behaviour_, so it's just about ergonomics. I was simply confused by the mix of types. That said, it's not really that important, and we've probably spent too long discussing it already :-)

ltratt: Ah, I think I see what you mean. @0152la maybe worth making this change, as I don't think it's a big one?

0152la: This should have been done in https://github.com/capablevms/cheri-examples/commit/1b46b9cebdf195bcedad7e24ee673a9450701388, as per the above, unless I'm misunderstanding?

jacobbramley: I think `1b46b9c` made them function pointers. I.e. so C thinks there's an 8-byte function pointer at `&switch_compartment_end`, which there isn't.

ltratt: I think @jacobbramley is suggesting `s/void */void/`.

0152la: I see; I was confused by the small detail. I think what we want is in https://github.com/capablevms/cheri-examples/pull/56/commits/420bddc50eba52cfd7d15f31e917ccdc38486d04.

ltratt: I think you also need to `s/extern void *comp_f_fn_end();/extern void comp_f_fn_end();/`?

0152la: I think they are as well [1]?

https://github.com/capablevms/cheri-examples/pull/56/files#diff-6d24dc24a1e7111722091fada0a3b1105a2b592c8eb46c5d327a3fda9ebf61c8R64-R65

ltratt: Oops, I must have been looking at the wrong thing -- sorry!

ltratt:
Please squash.

0152la:
Squashed.

PR #57, opened 2022-01-19
Add two negative examples

0152la:
This is to start a conversation, if there are any additional vectors of attack to be considered. I have in mind one compartment attempting to attack another, or attempting to modify its own capabilities somehow. Also, one issue that isn't solved in the `secure` example is checking the number of expected compartments. I plan on submitting a separate PR adding the check in, and a negative example to exercise that.

Adds to initial negative examples, expected to fail to show that the
security works. These contain:
* `redirect_clr` writes a capability to the `clr` pointing to some
  malicious code, in order for `switch_compartment` to jump to that code
  on return;
* `update_ddc` leaks the address of the DDC capability used by
  `switch_compartment`, allowing a container to attempt to load, modify,
  and update this capability.


ltratt:
Maybe silly question but is it possible to have these examples use the existing code? At the moment it's a bit hard to read the PR because I *think* there's a lot of code duplication, so I'm not sure where to focus my attention. Warning: I might be wrong!

0152la:
What I have done is mark the files changed (in relation to their counterparts in `secure/`) with a suffix. So `compartments-redirect_clr.s` is where changes are made compared to `compartments.s` (while `main.c` has no changes in `secure-redirect_clr` compared to `secure/main.c`). Thus, an option is doing a diff locally.

To help see things in the Github UI, I could remove this commit, force push one with just copies of the base `secure` files, and then run another commit with the changes over that?

Overall, yes, there is a ton of code duplication which I dislike, but it was the option I came up with to integrate stuff in our existing build workflow.

ltratt:
I think it'll benefit all of us to have a solution where we can clearly see the important deltas rather than having lots of repetition. Can you have a look into how to reuse code between examples? I don't know the best way to do it, but it must be possible to do something, whether that's e.g. moving Makefiles up a level or whatever. Feel free to force push an update if that's the best way of doing thngs.

ltratt:
Thanks!

0152la:
Ok, I've made a mess of everything, but for now, the differences should be emphasised in the latest commit [1], so this PR can proceed for now, at least in terms of correctness.

[1] https://github.com/capablevms/cheri-examples/pull/57/commits/6ca47879dad256f32220d2fc4d1f734c817a761a

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/switch_compartment-update_ddc.s
```diff
@@ -108,7 +108,8 @@ clean:
     mov x18, x18
     // x19-x29 are callee-saved, but only the lower 64 bits.
     mov x19, x19
-    mov x20, x20
+    // Simulate leaking adress of switcher DDC
+    //mov x20, x20
```
ltratt, line 112: Should this be deleted?

ltratt:
@jacobbramley Grateful for your thoughts on these examples!

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/switch_compartment-update_ddc.s
0152la: Actually no; I think the change on lines 54-56 should be reverted, and this change kept in to emphasise that we use that register to emulate the address leak.

ltratt: OK, maybe push a change for that (not force push as @jacobbramley might be reviewing as we speak)?

jacobbramley: You can force-push; I'm finishing something off and will probably get here after lunch.

0152la: Done in https://github.com/capablevms/cheri-examples/pull/57/commits/44bb3d8a3272205682b9d860f01a90b36e85092e.

0152la: I'd rather avoid force pushing, but those force pushes were done as we wanted the Github UI to comply, and that was the only way to do so.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/shared.S
```diff
@@ -0,0 +1,29 @@
...
+executive_switch:
+    mov      c29, c0
+    mov      x0, #0
+    cvtp     clr, lr
+    b        switch_compartment
+    ret      clr
+
+#include "switch_compartment.s"
+
+#include "compartments.s"
```
jacobbramley, line 29: This file doesn't exist.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
```diff
@@ -0,0 +1,169 @@
+/* This example attempts to set a `clr` which would jump to a malicious
```
ltratt, line 1: Could we highlight the line(s) where the error will happen?

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2021 The CapableVMs "CHERI Examples" Contributors.
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+SHARED_SOURCES := shared.s
+CFILES := $(wildcard *.c)
```
jacobbramley, line 5: I don't think these are correct. Using `Makefile.simple`:

- Each entry in `CFILES` maps onto a binary target with the same name (minus `.c`). Defaults to `*.c`.
- Each entry in `HFILES` is treated as a dependency, but not otherwise used. Defaults to `include/*.h`.
- Each entry in `SHARED_SOURCES` is treated as a dependency and passed to the compile command for _every_ binary target (as per `CFILES`).

I could, perhaps, have chosen better names, but in my defence, they worked for all the examples at the time.

To make this work as the `secure-redirect_clr` source files are written currently (reusing `Makefile.simple`):

```
SHARED_SOURCES := shared.S    # Note capital .S.
# CFILES := $(wildcard *.c)   # This is the default anyway. You can specify main.c if you prefer.
HFILES := $(wildcard *.h) $(wildcard ../../../include/*.h) $(wildcard ../../../../include/*.h) $(wildcard *.s)
    # ... or whatever else you used. Note that *.s is listed here because they're #included in shared.S.
```

This is a bit confusing because you have some `.s` files that are `#include`d, so they have to be included under `HFILES`, not `SHARED_SOURCES`.

For me, that causes `make` to try the following (which I think is correct):
```
[...]/clang --config cheribsd-morello-hybrid.cfg  main.c shared.S     -o bin/morello-hybrid/main
```

It doesn't build like that, because I lack `compartments.s`.

----

Whilst that would work, it's probably not ideal. It would be better to put all the .s (and .S) files into `SHARED_SOURCES` and drop the `#include`s. You'd have to declare the functions if you want to call them (like `.global switch_compartment`) but then everything works more conventionally, which is useful in an example!

I'd aim for this, and make the source files match:

```
SHARED_SOURCES := $(wildcard *.S)   # Just make all the files .S to keep this simple.
# CFILES := [default]
HFILES := $(wildcard include/*.h) $(wildcard ../../../include/*.h) $(wildcard ../../../../include/*.h)
     # Assuming that `main.h` is moved under `include`, like most other examples.
```

... so Make tries:

```
#                                     CFILES[0] ---v    /----------------- SHARED_SOURCES ----------------\
[...]/clang --config cheribsd-morello-hybrid.cfg main.c shared.S compartments.S compartments-redirect_clr.S -o bin/morello-hybrid/main
```

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2021 The CapableVMs "CHERI Examples" Contributors.
```
jacobbramley, line 1: Not 2022?

jacobbramley (commented):
I'm just making admin (Makefile) comments to start with. Also just for one example, assuming that the other follows a similar pattern.

Note that I can't build this locally (as noted in comments).

0152la:
The following is for buildbot testing

0152la:
bors try

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
0152la: Updated in https://github.com/capablevms/cheri-examples/pull/57/commits/e20156f8ebe99ccfd7a7bd113ff907bdc5d2ac92.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
0152la: I've added some reference to specific lines for both examples [1] [2]. Let me know if further clarification would be useful.

[1] https://github.com/capablevms/cheri-examples/pull/57/commits/c261cafc26a475cf8df2770c5afecb1b26f95093
[2] https://github.com/capablevms/cheri-examples/pull/57/commits/f62251051773440a4427a724ac22944980644c2e

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
ltratt: @0152la Has this comment been dealt with?

ltratt:
Broadly speaking this looks OK to me, but the details do elude me somewhat!

@jacobbramley can you do another check? If you're happy with it, I think we should get this merged, and then ask @0152la to write something high-level so that I have a better chance of understanding what's going on!

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
0152la: Not yet (that's why I changed it to a draft PR).

ltratt: Ah!

0152la:
Just note that I have still to work out the building issues; please do look at the actual code otherwise and see if there can be improvements done, or clarifying comments added.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/shared.S
0152la: Should be fixed in https://github.com/capablevms/cheri-examples/pull/57/commits/44440e75a75af8528307a00193052f3155aa9bed.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
0152la: This should be now fixed in https://github.com/capablevms/cheri-examples/pull/57/commits/44440e75a75af8528307a00193052f3155aa9bed. (with the corresponding changes for the second example done in https://github.com/capablevms/cheri-examples/pull/57/commits/4c9279a8a5e169b274b030a229d3f563f457043d)

0152la:
@jacobbramley The build system should have been fixed now, and `run-main` should work. I think the Makefiles as they are could be moved on level up and shared, but it's better to leave that for the next change. Do let me know if there are any further infrastructure issues, and your thoughts on the essential changes to the security of the switcher.

jacobbramley (commented):
Partial review because you added more updates; I'll come back and look at the finished whole.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
```diff
@@ -0,0 +1,175 @@
+/* This example attempts to set a `clr` (`compartments.S:12`) which would jump
```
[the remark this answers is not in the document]
0152la: I added them to try help @ltratt to better understand, but I do realise that absolute numbers are problematic. @ltratt , do you think they help as they are, and we should strive to maintain them, or leave the references more abstract, as proposed?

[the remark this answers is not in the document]
ltratt: An alternative to having the line numbers at the top is to add some sort of reference (maybe "look for XYZ in file ABC.c" or something?)?

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/compartments-redirect_clr.s
```diff
@@ -0,0 +1,42 @@
...
+ * appropriately to cover `switch_compartment`.
+ */
+.type comp_f_fn, "function"
+comp_f_fn:
+    // Set compartment ID we want to switch to
+    mov       x0, #1
+
+    // Try to set the CLR to jump to a malicious function
+    adr       x10, comp_f_malicious
+    cvtp      c30, x10
```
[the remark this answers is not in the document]
0152la: Amended in https://github.com/capablevms/cheri-examples/pull/57/commits/d9280d04c2f9b9a4eb76ec321d9d172e9f84dae9.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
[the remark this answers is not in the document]
0152la: Attempted to improve the wording in https://github.com/capablevms/cheri-examples/pull/57/commits/c7ba387c2cf4fbc31000173449a3d822f89700e2.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
```diff
@@ -0,0 +1,175 @@
...
+ * to a malicious function (`compartments.S:29`) after `switch_compartment`
+ * returns. The execution faults when entering the malicious function
+ * (`compartments.S:29`) via returning from `switch_compartment`
+ * (`switch_compartment.S:73`) as the bounds on the capability loaded into
+ * `clr` are inherited from the PCC of the compartment (`compartments.S:12`),
+ * being the only way the compartment itself can create new capabilities from
+ * scratch (assuming no access to other capabilities).
+ *
+ * Note that for brevity, filenames have been shortened to remove the common
+ * suffix which indicates that they are changed from the base `secure` example.
```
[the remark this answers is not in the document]
0152la: Changed in https://github.com/capablevms/cheri-examples/pull/57/commits/c7ba387c2cf4fbc31000173449a3d822f89700e2.

ltratt:
Once @jacobbramley is happy with this, you should squash.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/compartments-redirect_clr.S
```diff
@@ -0,0 +1,51 @@
...
+.global comp_g_fn_end
+
+.type comp_f_fn, "function"
+comp_f_fn:
+    // Set compartment ID we want to switch to
+    mov       x0, #1
+
+    // Try to set the CLR to jump to a malicious function
+    adr       x10, comp_f_malicious
+    cvtp      c30, x10
```
jacobbramley: Right, so this is the crux of the demonstration: the result (in `c30`/`clr`) is out of bounds. It would (hopefully) fail if you simply returned immediately with `ret clr`.

What this example shows, at least to me, is that the compartment switching mechanism doesn't break that guarantee.

Is my understanding correct?

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/main.c
```diff
@@ -0,0 +1,172 @@
+/* This example attempts to modify the DDC of `switch_compartment`, assuming it
+ * knows the address where it resides (leaked via `c20` in `shared.S:21`, and
```
jacobbramley: That file/line number doesn't seem to exist.

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/compartments-update_ddc.S
```diff
@@ -0,0 +1,53 @@
...
+    // Set compartment ID we want to switch to
+    mov       x0, #1
+
+    // Store the `clr` for exitting `comp_f_fn`; this is overwritten by
+    // `switch_compartment`.
+    str       clr, [sp, #-16]!
+
+    // Attempt to modify DDC of `switch_compartment`
+    mrs       c1, DDC
+    ldr       c2, [x20]
```
jacobbramley, line 25: What about using `c20`?

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/compartments-redirect_clr.S
0152la: That is correct. But to elaborate further, what this example is attempting to show is that the compartment, in some way, affects the environment that `switch_compartment` functions in. In this instance, it can affect the return address of the switcher. This is a sketch of how such an interaction could be subverted.

Further, this example can give rise to the following questions:
* what would be required for this approach to succeed in improperly modifying the return address;
* are there any other ways of deriving a `clr` such that the securities that work in this example are subverted/

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/main.c
0152la: I can't seem to find where this comment comes from. I'd say it comes from an outdated version, but Github doesn't mark it as such? I don't think this is how it is in the most current version [1], but please point me in the right direction if I'm wrong.

[1] https://github.com/capablevms/cheri-examples/pull/57/files#diff-2adcbeed62c5d5cd6351228f9eb71fd19efe139c901677c406515c94b78a02ffR1-R2

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
```diff
@@ -0,0 +1,175 @@
...
+	assert(id < COMP_COUNT);
+	struct comp new_comp;
+	new_comp.id = id;
+
+	new_comp.compartment_start = (void *) _start_addr;
+	new_comp.stack_addr = (void *) (_start_addr + comp_stack_size);
+	new_comp.stack_len = comp_stack_size;
+	new_comp.heap_len = total_comp_size - comp_stack_size;
+
+	// Ensure 16-byte alignment throught the compartment bounds
```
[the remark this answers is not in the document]
0152la: Fixed in https://github.com/capablevms/cheri-examples/pull/57/commits/fd8243747ddb5f45055e54fcd6510a05b0ab8019/ This typo probably appears in a few other files, so I'll probably fix those in due course.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/main.c
```diff
@@ -0,0 +1,175 @@
...
+void *__capability switcher_call;
+
+/* Abstract representation of a compartment. Within 80 bytes we represent:
+ * - an id (8B)
+ * - address to the start of the compartment memory area (also lowest address
+ *   of the stack) (8B)
+ * - address to the start (highest address) of the compartment's stack
+ *   (corresponding to the lowest address of the heap) (8B)
+ *   the size of the stack (8B)
+ * - address to the top (highest addressof the compartment's heap (8B)
```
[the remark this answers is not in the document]
0152la: Updated in https://github.com/capablevms/cheri-examples/pull/57/commits/20827d366a0c0b61b0750a0f2c7ab2fc3970d39f.

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/compartments-update_ddc.S
jacobbramley: What I mean is that I think the access will work if you use `c20`, since that ignores `DDC` and just uses the capability itself (which is now leaked due to changes in `clean` for this example).

Have I misunderstood?

0152la: What I attempted to do is store the address pointing to the DDC used by `switch_compartment` in `x20`, not the actual capability itself. `c0` in `executive_switch` (which is installed in `c20`) is the `ldpblr` capability we're using normally, so I'm just taking the address from that here.

ltratt:
@0152la Let's get this squashed and merged. I think we can improve this, probably, but I think now is the right time to do a slightly more radical rethink about how we structure things.

0152la:
Squashed.

ltratt:
bors r+

(If @jacobbramley has additional comments, we can address those in a subsequent PR.)

0152la:
This last commit should fix it; I can `try` until it works, if you want.

ltratt:
Good idea to `try` until it works.

0152la:
bors try

0152la:
bors try

ltratt:
Please squash.

0152la:
Squashed. For reference, the issue was that infrastructure changes affected previous examples, which I forgot to update.

PR #59, opened 2022-01-26
Add utility function and assertions to verify PCC bounds in shared objects

drpdr:
Draft PR to enhance what we already have in the shared_objects example. If you like the idea, I'll add the other tests/more tests we have discussed. Basically, I do with assertions what I checked in `gdb`.

### DIFF
--- a/shared_objects/include/find_sentries.h
+++ b/shared_objects/include/find_sentries.h
@@ -1,3 +1,8 @@
+/*
+ * Some utility functions to check range of addresses
+ * and whether a capability is the only one in the PCC
+ */
+
 #include <stdbool.h>
 
 #include "../../include/common.h"
@@ -38,4 +43,16 @@ bool scan_range(void *ptr, void *exact)
 		}
 	}
 	return found;
+}
+
+/**
+ * Check whether ptr1 and ptr2 share the same base and
+ * bounds. If the latter is true and, e.g., ptr1 is the PCC,
+ * then the PCC contains only ptr2.
+ */
+bool base_and_bounds_eq(void *__capability ptr1, void *__capability ptr2)
+{
+	void *__capability pcc = cheri_pcc_get();
+	return cheri_base_get(ptr1) == cheri_base_get(ptr2) &&
+		   cheri_length_get(ptr1) == cheri_length_get(ptr2);
 }
--- a/shared_objects/static_function.c
+++ b/shared_objects/static_function.c
@@ -1,0 +1,39 @@
+#include "include/static_function.h"
+#include "include/find_sentries.h"
+#include <assert.h>
+#include <stdio.h>
+
+static int static_function()
+{
+	assert(base_and_bounds_eq(cheri_pcc_get(), &static_function));
+	return 0;
+}
+
+static double another_static_function()
+{
+	assert(base_and_bounds_eq(cheri_pcc_get(), &another_static_function));
+	return 1.0;
+}
+
+void non_static_fun()
+{
+	assert(base_and_bounds_eq(cheri_pcc_get(), &non_static_fun));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &static_function));
+	printf("Called static function with result: %d\n", static_function());
+}
+
+void non_static_fun2()
+{
+	assert(base_and_bounds_eq(cheri_pcc_get(), &non_static_fun2));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &another_static_function));
+	printf("Called another function with result: %f\n", another_static_function());
+}
+
+void check_static_and_non_static()
+{
+	assert(base_and_bounds_eq(cheri_pcc_get(), &non_static_fun));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &static_function));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &non_static_fun2));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &another_static_function));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &check_static_and_non_static));
+}
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -8,6 +8,7 @@ set -e
 
 export CHERIBASE=~/cheri
 CHERIBUILD=~/build
+export TEST_TARGET=$1
 
 echo "Checking clang-format..."
 # Note that `sdk` and `morello-sdk` contain different versions of clang-format,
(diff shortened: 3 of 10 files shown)


shared_objects/include/find_sentries.h
```diff
@@ -39,3 +44,12 @@ bool scan_range(void *ptr, void *exact)
 	}
 	return found;
 }
+
+bool is_pcc_exposed(void *__capability ptr)
```
ltratt, line 48: Let's add a documentation comment for this function (the name doesn't quite make sense to me -- I think this function is checking "is ptr contained within the PCC's bounds?" but I might be wrong!).

drpdr: Yes, I was unsure about the name. This functions is checking that `ptr` is the only one contained in the `PCC`.

ltratt: What do you mean by "the only one"?

drpdr: You are right. Let me rephrase. I mean the PCC's bounds and ptr's are the same. The PCC doesn't cover something else than ptr.

drpdr: We generally don't want PCC's bounds over the ones of a certain function when executing it from a shared object. We want to ensure to have only that function in the PCC and no possibility for anything else.

ltratt: OK, I see: you're looking at defining some form of equality between `pcc` and `ptr`. You're going to have to nail down precisely what you want that equality to be, as there are several potential notions: in this case it looks like you want PCC + `ptr` to share the same base and bounds.

If I've got that right then I think you can simplify the code to be just:

```
cheri_base_get(pcc) == cheri_base_get(ptr) && cheri_length_get(pcc) == cheri_length_get(ptr)
```

The problem then becomes finding a good name for this: `pcc_and_ptr_base_and_bounds_eq` or something similar might have to do.

drpdr: Correct. Thanks. What about `pcc_shares_base_bounds_with(...)`? Too long, maybe.

ltratt: `pcc_base_and_bounds_eq_to` would work for me.

drpdr: There was also a subtle bug I hadn't noticed. If the PCC is accessed inside the utility function it loses its meaning. In 869072f.

ltratt: Broadly OK for me. Though `functionPtr` is not an idiomatic C variable name IMHO and `== true` is redundant. So let's change:

```c
void (*functionPtr)() = &increment;
assert(base_and_bounds_eq(cheri_pcc_get(), functionPtr) == true);
```

to:

```c
assert(base_and_bounds_eq(cheri_pcc_get(), &increment);
```

drpdr: Oh, yes switching between C & C++ and naming conventions... I'l fix the assertion too.

drpdr: Fix in 7c66820.

ltratt: Works for me! Please carry on with the rest of the PR.

drpdr:
In cf3ac22 I have added some checks for static functions. There is always only one function covered, i.e. they share the same bounds and no more. What we want to ensure is that there are "no situations where the PCC covers more than one function at a certain point in the program".

ltratt:
Can we alter the examples so that they don't need user input? I can't see any advantage to this, and a big disadvantage: we can't run these as automatic tests.

drpdr:
> Can we alter the examples so that they don't need user input? I can't see any advantage to this, and a big disadvantage: we can't run these as automatic tests.

Do you want this for all of them or only for the ones of this PR?

ltratt:
Let's only change the ones affected by this PR.

drpdr:
@ltratt I had forgotten about [this](https://github.com/capablevms/cheri-examples/blob/1efbed872bc36430a2a0d01bcdad8d9cee53913f/.buildbot.sh#L57). The examples are tested only on `morello-hybrid`. So, we cannot add a `riscv64-purecap` qemu instance. I can add the tests but not run them with `buildbot`.

drpdr:
> Let's only change the ones affected by this PR.

In 626156e.

shared_objects/static_function.c
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void check_static_and_non_static()
+{
+	assert(base_and_bounds_eq(cheri_pcc_get(), &non_static_fun));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &static_function));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &non_static_fun2));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &another_static_function));
+	assert(base_and_bounds_eq(cheri_pcc_get(), &check_static_and_non_static));
+	printf("Static and non static checks...\n\n");
```
ltratt, line 39: Let's only have `print`s in tests if they're important (e.g. to help debug likely problems). I don't know which are and aren't important so I will leave the selection to you :)

drpdr: They probably are not important at all. We can remove them.

ltratt: Excellent!

drpdr:
bors try

drpdr:
bors try

shared_objects/static_function.c
drpdr: Fix in 91f2d41.

drpdr:
Test for `riscv64-purecap` in 841ef77. We are going to run tests for the architecture we have been using, i.e. `morello-hybrid`, `riscv64-purecap`. I am going to run tests for all the architecture (at least put them here) in a subsequent PR.

drpdr:
> `.buildbot.sh` isn't called with arguments I don't think so what value will `$1` have?

You won't see `.buildbot.sh` called anywhere in here (with or without arguments). It is part of the infrastructure of `buildbot`. I pass the arg there.

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
This should be ready for squashing, I think.

ltratt:
Please squash.

drpdr:
Squashed.

drpdr:
Interesting, this worked multiple times yesterday and we haven't touched it...

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try-

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try

drpdr:
bors try

drpdr:
OK, @ltratt. Whenever you can, please, you can retry to merge.

PR #67, opened 2022-08-16
Add morello-purecap and fix naming

drpdr:
We had some renaming and we forgot about this repository.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -40,7 +40,7 @@ done
 export SSHPORT=10021
 export PYTHONPATH="$CHERIBUILD"/test-scripts
 
-if [ "$1" = "riscv64" ]; then
+if [ "$1" = "riscv64-purecap" ]; then
     echo "Running tests for 'riscv64-purecap' using QEMU."
     args=(
         --architecture riscv64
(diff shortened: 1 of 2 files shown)


.buildbot.sh
```diff
@@ -71,8 +71,6 @@ elif [ "$1" = "morello-hybrid" ]; then
     BUILDBOT_PLATFORM=morello-hybrid python3 tests/run_cheri_examples.py "${args[@]}"
 else
     echo "$1 not supported (yet)."
-    exit 1
-fi
-
-# TODO: Run tests for `morello-purecap` and `riscv64-purecap` too
-# This is related to https://github.com/CTSRD-CHERI/cheribuild/issues/182
+    # Exit gracefully
+    exit 0
```
ltratt, line 75: `exit 0` isn't needed but, also, I think `exit 1` was better (it is an error condition after all).

drpdr: We don't want the build to fail but we want it to continue with the other platform. Like [here](https://github.com/capablevms/llvm-function-split/blob/8ea6c0ca58c29a7a1f9d7519f31bb76a60940789/.buildbot.sh#L54).

ltratt: If we don't want a command calling this to fail, we need to signal that another way IMHO (or check that the platform is supported before calling this). Put another way: commands that say they've succeeded but haven't done the expected work are dangerous.

drpdr: Well, that was not correctly phrased. It's technically not a failure, we just want to inform the user. After several discussions during the past months, we agreed on passing the "platform" to `buildbot` because `cheribuild` has that `SIGT*" bug preventing multiple terminals.

ltratt: I think now is a good time to rethink this, unless it's really difficult to do so. Imagine if I did (say) "sh -c 'X'" and `X` failed but `sh` succeeded: we'd be bamboozled.

drpdr: After adding `morello-purecap` support (currently missed), we can remove that branch if you wish. It is not needed in this repository. I can add `morello-purecap` in this PR or in a subsequent PR.

ltratt: Whichever you think best.

drpdr:
bors try

drpdr:
bors try

drpdr:
@ltratt this is now ready for review.

ltratt:
Please squash.

drpdr:
> Please squash.

Done.

PR #70, opened 2022-10-14
Add an example that reads a capability from a file.

gabi-250:
According to the CHERI C/C++ programming guide:

> In the CheriABI process environment, only untagged data (not tagged
> pointers) may be written to or read from files. If a region of memory
> containing valid pointers is written to a file, and then read back, the
> pointers in that region24 will no longer be valid.

However, as this example demonstrates, this behaviour appears to be broken, as it is possible to read a valid capability from a file.

This example writes a capability to a temporary file and then successfully retrieves the original (valid) capability, which should not be possible according to the CheriABI.

The problem still reproduces if we simplify the example further by removing the part where the capability is written to disk and hard-coding the file from which to read the capability (the file could contain `ff5c fff7 ffff 0000 ff5c 7f60 4000 dc5d` for example).


ltratt:
I am tentatively calling this the "hidden buffer" problem. I *think* (but @gabi-250 may have been able to look?) that CheriBSD's libc probably has a read/write buffer and that you can "read" previously written values from that buffer without going via the disk. [If there is no such buffer, then we would seem to have a much deeper problem on our hands!]

@jacobbramley I'd welcome your thoughts on this example!

bors try

gabi-250:
@ltratt I have not dug into CheriBSD's libc yet, because I managed to reproduce the issue even with buffering disabled, by using:
* `read`/`write` (which are just thin wrappers around the `read`/`write` syscalls -- I double-checked this is indeed the case) instead of `fread`/`fwrite`
* by disabling buffering using `setvbuf` (`setvbuf(f, NULL, _IONBF, 0 )`)
* by removing the part where the capability is written to disk and hard-coding the file from which to read the capability (the file could contain ff5c fff7 ffff 0000 ff5c 7f60 4000 dc5d for example).

ltratt:
It *might* be that the kernel is doing the buffering?

gabi-250:
> Duh, I ran bors, but the new file isn't linked to the build. @gabi-250 can you push a commit that links the new commit to the build in one way or another? Thanks!

I added `cap_to_file` to `tests/run_tests.sh` in [72aa8a9](https://github.com/capablevms/cheri-examples/pull/70/commits/72aa8a99227282a116f5407f926bfb6d7e9d095e)

ltratt:
Thanks!

bors try

jacobbramley:
Agreed, a file I/O buffer sounds like the most likely culprit for the behaviour in the patch. This should be reported as a CheriBSD bug.

I was considering suggesting a few other experiments (e.g. avoiding `tmpfile()`, which has some special behaviours), but this concerns me:

> The problem still reproduces if we simplify the example further by removing the part where the capability is written to disk and hard-coding the file from which to read the capability (the file could contain ff5c fff7 ffff 0000 ff5c 7f60 4000 dc5d for example).

The I/O buffer can't explain that, can it? If I understand correctly, that's simple capability forging, and definitely shouldn't be possible.

ltratt:
>> The problem still reproduces if we simplify the example further by removing the part where the capability is written to disk and hard-coding the file from which to read the capability (the file could contain ff5c fff7 ffff 0000 ff5c 7f60 4000 dc5d for example).
> The I/O buffer can't explain that, can it? If I understand correctly, that's simple capability forging, and definitely shouldn't be possible.

Agreed, but I'm not quite sure I understood the example exactly. @gabi-250 do you think you could produce a second file with our hard-coded suggestion (and link it to the build) so we can have a look? Thanks!

gabi-250:
@jacobbramley 

> I was considering suggesting a few other experiments (e.g. avoiding tmpfile(), which has some special behaviours), but this concerns me:

I tried avoiding `tmpfile()` too, but it did not help.

> The I/O buffer can't explain that, can it? If I understand correctly, that's simple capability forging, and definitely shouldn't be possible.

It's possible I'm doing something wrong here (I'm still very new to CHERI).

ltratt:
I love bors!

```
Assertion failed: (!cheri_is_valid((void *) stored_cap)), function main, file cap_to_file.c, line 93.
```

which confirms the basic bug.

gabi-250:
> Agreed, but I'm not quite sure I understood the example exactly. @gabi-250 do you think you could produce a second file with our hard-coded suggestion (and link it to the build) so we can have a look? Thanks!

Which hard-coded suggestion? The simplified version of this example (described in the commit message)?

cap_to_file.c
```diff
@@ -0,0 +1,98 @@
...
+	if (read_cap(f, &stored_cap))
+	{
+		return 1;
+	}
+
+	printf("Read capability from file: %#lp\n", (void *) stored_cap);
+	assert(cap == stored_cap);
+
+	// NOTE: This assertion should not fail, but it does
+	assert(!cheri_is_valid((void *) stored_cap));
```
ltratt, line 93: FWIW, I would suggest this becomes:

```c
// XXX the capability should not be valid but due to a bug in CheriBSD
// this assertion succeeds!
assert(cheri_is_valid((void *) stored_cap));
```

so that bors can pass.

ltratt:
> Which hard-coded suggestion? The simplified version of this example (described in the commit message)?

Yep!

cap_to_file.c
gabi-250: Done, see [8b3c77a](https://github.com/capablevms/cheri-examples/pull/70/commits/8b3c77a6be66c2281dc5e4c39626d62c0e91a622).

ltratt:
bors try

gabi-250:
> > Which hard-coded suggestion? The simplified version of this example (described in the commit message)?
> 
> Yep!

I pushed the simplified example in [8b3c77a](https://github.com/capablevms/cheri-examples/pull/70/commits/8b3c77a6be66c2281dc5e4c39626d62c0e91a622). Perhaps unsurprisingly, this does not reproduce the issue I described (the capability is indeed invalid). However, I _can_ manually trigger the bug by writing the file to disk and re-reading it (this seems to point to a kernel buffering issue as suggested by @ltratt )

ltratt:
>I pushed the simplified example in [8b3c77a](https://github.com/capablevms/cheri-examples/pull/70/commits/8b3c77a6be66c2281dc5e4c39626d62c0e91a622). Perhaps unsurprisingly, this does not reproduce the issue I described (the capability is indeed invalid). However, I can manually trigger the bug by writing the file to disk and re-reading it (this seems to point to a kernel buffering issue as suggested by @ltratt 

Right, got it! So the `bors try` should fail (I think).

@jacobbramley speculated that it's probable that we can set up an example where we `fwrite` a capability in one process and `fread` it in another and receive a valid capability in that second process. I suspect this will work too. Maybe we can create a simple example for that?

[@jacobbramley also speculated that this problem might exist with pipes. I think that's possible too! But one thing at a time etc etc :)]

gabi-250:
> Right, got it! So the bors try should fail (I think).

I expect it to pass, because the assertion in the second example asserts the right thing (I left both examples in the PR ) :)

ltratt:
> I expect it to pass, because the assertion in the second example asserts the right thing (I left both examples in the PR ) :)

Oops, I'm with you!

It did fail but not for a deep reason:

```
mkdir: read-cap-from-file: File exists
```

gabi-250:
bors try

gabi-250:
> @jacobbramley speculated that it's probable that we can set up an example where we fwrite a capability in one process and fread it in another and receive a valid capability in that second process. I suspect this will work too. Maybe we can create a simple example for that?

Sounds good, I can come up with an example and add it to this PR.

gabi-250:
Ready for another review.

gabi-250:
> Agreed, a file I/O buffer sounds like the most likely culprit for the behaviour in the patch. This should be reported as a CheriBSD bug.

I'll prepare a bug report.

ltratt:
Let's hold off on a bug report until we've done a bit more investigation of whether this is a one-off bug or there's something a bit more systemic going on.

gabi-250:
Yep, just gathering data for now :)

gabi-250:
I split `leak_capability.c` into two components as suggested (a reader and a writer). As expected, the behaviour is the same.

gabi-250:
bors try

leak-capability/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -eou pipefail
+
+# Write the capability
```
gabi-250, line 5: I've not added a loop because it always seems to succeed on the first try.

leak-capability/build/Makefile.leak-capability
```diff
@@ -0,0 +1,39 @@
...
+
+RUN_SH := ./run.sh
+
+include ../build/Makefile.simple
+
+.PHONY: scp-run-sh
+
+run-leak-capability: $(BINS) scp-run-sh
+ifdef RUNDIR
+	scp $(SCP_OPTIONS)  -P $(SSHPORT) $(BINS) $(RUNUSER)@$(RUNHOST):$(RUNDIR)
```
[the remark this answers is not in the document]
gabi-250: Fixed

ltratt:
I think we just have the "rename and plonk in a directory" fix to do then we'll be ready to merge!

gabi-250:
I moved all the new examples to `capability_sharing` in [5825013](https://github.com/capablevms/cheri-examples/pull/70/commits/5825013ebe06d9f6fcf04225de71480b8b8b1b51)

gabi-250:
bors try

gabi-250:
bors try

ltratt:
Please squash.

gabi-250:
Squashed (one commit per example).

PR #71, opened 2022-10-17
Add examples that write capabilities to `mmap`ed memory.

gabi-250:
Add examples that write capabilities to `mmap`ed memory.

These examples demonstrate what happens when you write tagged pointer values to various types of memory mappings:
   * Anonymous private mapping (OK)
   * Anonymous shared mapping (OK)
   * File-backed private mapping (OK)
   * File-backed shared mapping (segfault)

### DIFF
--- a/capability_sharing/cap-to-file/README.md
+++ b/capability_sharing/cap-to-file/README.md
@@ -1,6 +1,6 @@
 # Read a capability from a file
 
-This example is not automatically built by the top-level makefiles, but are
+This example is not automatically built by the top-level makefiles, but is
 built in a similar manner:
 
 ```
--- a/capability_sharing/leak-capability/README.md
+++ b/capability_sharing/leak-capability/README.md
@@ -19,7 +19,7 @@ it is possible to read a valid capability from a file.
 
 ## Running
 
-This example is not automatically built by the top-level makefiles, but are
+This example is not automatically built by the top-level makefiles, but is
 built in a similar manner:
 
 ```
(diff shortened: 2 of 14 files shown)


gabi-250:
> > File-backed shared mapping (segfault)
> 
> Is this the expected semantics do you think? I hadn't really thought about this use case to be honest.

I expected it to receive a SIGPROT signal to be honest, although that doesn't make much sense in hindsight, since it wouldn't really a protection violation. The CHERI docs don't explicitly state what should happen in this case (so I wrote these examples to try it out for myself):

> If a region of memory containing valid pointers is written to a file, and then
> read back, the pointers in that region will no longer be valid. If a file is
> memory mapped, then pages mapped copy-on-write (MAP_PRIVATE) are able to hold
> tagged pointers, since they are swap-backed rather than file-backed, but pages
> mapped directly from the buffer cache (MAP_SHARED) are not.

In other words, `MAP_SHARED` pages aren't able to hold capabilities (the docs don't explain what "aren't able to hold capabilities" means in practice though).

Anyway, this was a bit surprising to me, but it's definitely [not a bug](https://github.com/CTSRD-CHERI/cheribsd/blob/6837f8e653d9f3edb4022acb69d1bd11fb762c2f/bin/cheribsdtest/cheribsdtest.c#L877) (see also [this CheriBSD test](https://github.com/CTSRD-CHERI/cheribsd/blob/4511a9b9d97ab3ca8041873e353d818d1541d8fc/bin/cheribsdtest/cheribsdtest_vm.c#L433-L452))

ltratt:
Got it. Then i think this is ready to merge?

gabi-250:
Sure.

PR #72, opened 2022-10-17
A couple of examples which fail to leak capabilities using pipes.

gabi-250:
Interestingly enough, I can't reproduce the issue highlighted in my [previous PR](https://github.com/capablevms/cheri-examples/pull/70) using Unix pipes for IPC.

This PR contains 2 examples, both of which seem to work correctly (as in, any capabilities read from a pipe are going to be invalid).

The first example consists of two sibling processes connected to a pipe:
   * a writer process that writes a capability to the write end of the pipe, and
   * a reader process that reads the capability written by the writer process to the pipe. The reader reads an invalid capability

The second example is similar, except it relies on the shell to connect the two processes to a pipe.


ltratt:
Out of curiosity, if you create a pipe and read/write to it in the same process, does it still (correctly) invalidate the capabilities? [I would guess "yes" but I don't know.]

gabi-250:
Yep, the capability is still invalidated!

PR #74, opened 2022-10-28
Add the missing `base.h` file.

gabi-250:
The `comp_setup` example doesn't build on `master`:
```
gabi@bencher14 ~/c/h/c/comp_setup > SSHPORT=10085 make -f ../Makefile.morello-hybrid run-main
/home/gabi/cheri/output/morello-sdk/bin/clang -E  shared.S > shared.s
shared.S:4:10: fatal error: 'main.h' file not found
         ^~~~~~~~
1 error generated.
make: *** [<builtin>: shared.s] Error 1
rm shared.s
```

It looks like `main.h` was accidentally removed at some point (#57). This patch reinstates it as `base.h` (for consistency with the other examples).

PR #75, opened 2022-10-31
Refactor switch_compartment.s

gabi-250:
This PR contains some changes that helped me grok the code a little better. If you disagree with this refactoring feel free ~to decline this change~ to let me know why. Otherwise, I might follow up with a couple more dedupings.

* Refactor `switch_compartment.s`
As the diffs below show, the `switch_compartment.s`s of the 3 `secure` examples (`secure`, `secure-redirect_clr`, `secure-try_deref`) are almost identical:
```
gabi@areca ~/c/cheri-examples > diff hybrid/compartment_examples/inter_comp_call/secure/switch_compartment.s hybrid/compartment_examples/inter_comp_call/secure-try_deref/switch_compartment.s
38c38
<     // Save old DDC (c2), old SP (x12), old CLR (clr) on stack
---
>     // Save old DDC (c2), old SP (x12), old LR (clr) on stack
gabi@areca ~/c/cheri-examples > diff hybrid/compartment_examples/inter_comp_call/secure/switch_compartment.s hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/switch_compartment.S
5a6,8
>
> #include "../include/secure.h"
>
6a10,11
> .global switch_compartment_end
>
```

Since the only real difference between them is the exported `switch_compartment_end` symbol, it's safe to refactor the 3 examples to use the same `switch_compartment.s`. Deduplicating the code makes it easier to understand the differences between the `secure_*` variants.

* Conditionally include a `clean` implementation.

Some `inter_comp_call` examples use a correct `clean` implementation, whereas the `secure-update_ddc` example uses one that intentionally fails to clear `x20`. We can conditionally include one of the two implementations in `switch_compartment.S` to further deduplicate the code.


hybrid/compartment_examples/inter_comp_call/clean_leak_ddc.s
```diff
@@ -0,0 +1,41 @@
...
+    mov x15, #0
+    mov x16, #0
+    mov x17, #0
+    // x18 is the "platform register" (for some platforms). If so, it needs to
+    // be preserved, but here we assume that only the lower 64 bits are
+    // required.
+    mov x18, x18
+    // x19-x29 are callee-saved, but only the lower 64 bits.
+    mov x19, x19
+    // Simulate leaking address of switcher DDC
```
0152la, line 28: If I read the tree correctly, this will be going in the "root" folder, rather than in one specific example directory. However, this comment, I believe, corresponds to one of the particular examples, where we chose `c20` to be the target for DDC switching. I would perhaps rephrase this comment to something like "simulate leaking a register, used to evaluate security issues".

hybrid/compartment_examples/inter_comp_call/clean.s
```diff
@@ -0,0 +1,49 @@
+// Inner helper for cleaning capabilities from registers, either side of an
```
0152la, line 1: These `.s` files are now an indirect dependency in the `Makefile`, since they are included by `switch_compartment.S`, and they are essentially "hidden" from the compilation process. While I appreciate the de-duplication, perhaps there is a better way to go around this, rather than hiding this process from the reader...

hybrid/compartment_examples/inter_comp_call/secure-try_deref/shared.S
```diff
@@ -1,8 +1,6 @@
 // Copyright (c) 2021 The CapableVMs "CHERI Examples" Contributors.
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-#include "../include/secure.h"
```
0152la, line 4: Was this an unneeded include?

hybrid/compartment_examples/inter_comp_call/switch_compartment.S
```diff
@@ -72,54 +78,14 @@ switch_compartment:
...
-    mov x27, x27
-    mov x28, x28
-    mov x29, x29  // FP
-    // We need LR (x30) to return. The call to this helper already cleaned it.
-    // Don't replace SP; this needs special handling by the caller anyway.
-    ret
+// Select the appropriate `clean` implementation. Some examples use a correct
+// `clean` implementation, whereas `secure-update_ddc` uses one that
+// intentionally fails to clear `x20`.
+#ifdef INTER_COMP_LEAK_DDC
```
0152la, line 84: Similar to another comment above, there is a dependency here from within a specific example (namely this `INTER_COMP_LEAK_DDC` variable). But perhaps the readability is improved by the two distinct header file names.

hybrid/compartment_examples/inter_comp_call/secure/compartments.s
```diff
@@ -30,7 +30,6 @@ comp_f_fn_end:
  */
 .type comp_g_fn, "function"
 comp_g_fn:
-    mrs       c10, DDC
```
0152la, line 33: Is this intentional?

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/shared-update_ddc.S
```diff
@@ -13,3 +13,6 @@ executive_switch:
     cvtp     clr, lr
     b        switch_compartment
     ret      clr
+
+#define INTER_COMP_LEAK_DDC 1
```
0152la, line 17: If we do keep this, I think a comment is in order, as to where it's used.

0152la (commented):
I think overall, I prefer having the original structure, where each test is within each folder, and changes to files are then emphasized by file suffixes, so one can easily tell where the changes for each test is introduced. As such, I would argue duplicating this code in completely distinct files helps someone know "oh, there's definitely not anything interesting happening in these files, the changes are in these files which I can put side by side". I think having top-level files with conditional paths makes it harder to see the logic, than just diffing the folders. However, that's perhaps just how I approach things, and it might be better this way, so I won't give a strong decline, but I am partly against it.

Granted, there are perhaps some slight syntactical differences when diffing some files that might take away attention from the logic. Those are definitely worth fixing.

gabi-250:
> I think overall, I prefer having the original structure, where each test is within each folder, and changes to files are then emphasized by file suffixes, so one can easily tell where the changes for each test is introduced. 

I can rework this to be more to your liking, but i would argue the shared code needs to be refactored out of the examples themselves.

> As such, I would argue duplicating this code in completely distinct files helps someone know "oh, there's definitely not anything interesting happening in these files, the changes are in these files which I can put side by side". I think having top-level files with conditional paths makes it harder to see the logic, than just diffing the folders. However, that's perhaps just how I approach things, and it might be better this way, so I won't give a strong decline, but I am partly against it.

De-duplicating the code is meant to save you the trouble of having to manually diff the files. To someone new to the codebase it's not obvious at all which files are interesting and which aren't, which is why I find this version of the code easier to understand (with boilerplate out of the way, I can focus on the parts that _are_ different).

hybrid/compartment_examples/inter_comp_call/secure/compartments.s
gabi-250: Good catch, I actually meant to submit this as a separate PR - it looks like this instruction isn't needed at all (`c10` is never referenced again in `comp_g_fn`).

hybrid/compartment_examples/inter_comp_call/clean.s
gabi-250: My understanding is they can't really be compiled separately as they need to in the `[switch_compartment, switch_compartment_end)` address range, which is why they're `#include`d in `switch_compartment.S`. Also, plenty of other `.S` files `#include` other assembly sources, so perhaps you are suggesting this is a more general problem in this project?

gabi-250:
Anyway, I'm not married to this change, so feel free to decline it if you disagree.

ltratt:
I am generally a fan of factoring out the common stuff and this repo could probably do with a bit of such tidying up!

A question then is: where do we put the common stuff? I don't feel too strongly about that, and I tend to feel that once we've factored out the common stuff, moving it later then becomes easier.

So, personally, I tend to think this PR nudges things in a slightly better direction, even if I'm sure it could be done even better in the future.

0152la:
Sure, I agree with an overall refactoring of the whole repo (particularly lots of `Makefile`s lying around and horrible relative paths all over the place, particularly in this section of the repo), so let's get this in as a first step towards that direction then.

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2022 The CapableVMs "CHERI Examples" Contributors.
 # SPDX-License-Identifier: MIT OR Apache-2.0
 
-SHARED_SOURCES := $(wildcard *.S)
+SHARED_SOURCES := $(wildcard *.S) ../switch_compartment.S
```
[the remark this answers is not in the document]
0152la: And maybe further, I think moving these makefiles out of each individual example folder would be a good goal, although I'm sure I must've tried this and given up for some reason or another...

hybrid/compartment_examples/inter_comp_call/clean.s
0152la: > as they need to in the [switch_compartment, switch_compartment_end) address range,

Right, I think I forgot that this example tries to do some execution compartmentalization (I did forget)? If that is the case, then this make sense, and they can't just be stand-alone files.

> Also, plenty of other .S files

Perhaps. I didn't appreciate the difference between `.s` and `.S` when I initially started working in this repo, so it might be worth considering. On the other hand, if it works, don't fix it.

0152la:
One other thing, as indicated by #74 , can we ensure these tests are run in the CI? They might already be, but please double check.

gabi-250:
> One other thing, as indicated by #74 , can we ensure these tests are run in the CI? They might already be, but please double check.

The `inter_comp_call` ones already run in CI (I can add the missing ones we talked about in #74, but it's probably best to do that in a separate PR).

gabi-250:
> A question then is: where do we put the common stuff? I don't feel too strongly about that, and I tend to feel that once we've factored out the common stuff, moving it later then becomes easier.

This PR scatters common stuff (`clean.s`, `clean_leak_ddc.s`, `switch_compartment.S`) in the `inter_comp_call` directory:
```
gabi@areca ~/c/c/h/c/inter_comp_call > tree -L 1
.
├── base
├── clean_leak_ddc.s
├── clean.s
├── include
├── secure
├── secure-redirect_clr
├── secure-try_deref
├── secure-update_ddc
└── switch_compartment.S
```

Perhaps it would be better to move them to a `common/` subdir.

Also (and this probably doesn't shouldn't be addressed in this PR), maybe we can sprinkle some e.g. `-Iinclude/`s in the Makefiles so we don't have to write relative paths when `#include`ing headers.

hybrid/compartment_examples/inter_comp_call/clean.s
gabi-250: > Perhaps. I didn't appreciate the difference between .s and .S when I initially started working in this repo, so it might be worth considering. On the other hand, if it works, don't fix it.

Oh, I personally don't mind unnecessary uses of `.S` (`.S` just means "run the pre-processor on this file"). What I meant to say is there are other `.s``/.S` files in this project that are "hidden" from the compilation process (i.e. they are inlined in a main `.s`/`.S` rather than compiled individually).

0152la: Right, I guess once we do a major refactoring, we can maybe care if this is a problem or not. For now, this change is fine. The other part, regarding the location of `clean` needing to be within the specified range, if this is due to execution compartmentalization, then we can leave this as is.

ltratt:
@gabi-250 A quick question based on our in-person conversation: I *think* you're still making further changes to this PR?

gabi-250:
Yep, I've addressed the comments but I haven't pushed yet because I realized the `inter_comp_call/base` example also could use a similar refactoring (otherwise it'll be the only example left in `inter_comp_call` that has its own Makefile - the others will be refactored as per [Andrei's comment](https://github.com/capablevms/cheri-examples/pull/75#discussion_r1010359738))

hybrid/compartment_examples/inter_comp_call/switch_compartment.S
gabi-250: Good point, though now I'm a bit unsure whether conditionally including a `clean` implementation is such a good idea. As you pointed out, it does hurt readability.

Maybe it would be better to have two `clean` functions (`clean_all` and `clean_leak_x20`): the compartment switcher can just choose one or the other based on the value of a (new) argument passed by the caller (i.e. `switch_compartment` receives an additional `leak_x20` arg).

hybrid/compartment_examples/inter_comp_call/clean.s
gabi-250: Yeah, `clean` needs to be within the bounds of the PCC of the compartment switcher (otherwise calling `clean` would trigger a CHERI protection violation).

hybrid/compartment_examples/inter_comp_call/switch_compartment.S
0152la: Some thoughts:
* Could lower the `#define` on the specific register we leak. That might allow to lift `clean` back up into `switch_compartment`, if you prefer;
* We could forego entirely the cleaning in the example where we want to leak. Should still keep the essence of what we want to showcase (i.e., capability leak at the register level), without fumbling about with conditionals;
* Two distinct functions are not bad, but there might be a better way than having to pass "state" (i.e., whether we want to leak or not) around. It might be easier for me to evaluate the idea with a concrete example, if you choose this path.

gabi-250: > Could lower the #define on the specific mov instruction for the register we want leak, inside clean itself. That might allow to lift clean back up into switch_compartment, if you prefer (although would require preprocessing);

Good point, though I think want to get rid of preprocessing altogether: ideally, all examples should assemble `switch_compartment.S` and link it in the final binary. Currently, the examples `#include` the top-level `switch_compartment.S` from their individual `shared.S`s. In most cases the `#include` isn't necessary (it could just as well be compiled separately and then linked in). However, this conditional include to select the appropriate `clean` makes this difficult to get rid of (someone needs to define `INTER_COMP_LEAK_DDC`: it's either going to be a `#define` preceding the `#include switch_compartment.S`, or a flag passed to the assembler. I think both these options would be less readable than removing preprocessing and adding a `leak_x20` arg).

> We could forego entirely the cleaning in the example where we want to leak. Should still keep the essence of what we want to showcase (i.e., capability leak at the register level), without fumbling about with conditionals;

True, but we'd still need a conditional to skip calling `clean` in `switch_compartment`

> Two distinct functions are not bad, but there might be a better way than having to pass "state" (i.e., whether we want to leak or not) around. It might be easier for me to evaluate the idea with a concrete example, if you choose this path.

Let me try to come up with an example.

0152la: Those were just some ideas that might help; feel free to commit what you think is best, and we can continue iterating over that, rather than preparing examples and so.

gabi-250:
This refactoring is proving a bit trickier than I anticipated. I'm debugging an issue where the refactored versions of the examples are `SIGPROT`-ing at an unexpected point, so I don't have anything new ready for review at this point. Anyway, in the interest of transparency I'm going to push some WIP changes (which you are free to ignore).

hybrid/compartment_examples/inter_comp_call/secure/compartments.s
gabi-250: Fixed in [642eac5](https://github.com/capablevms/cheri-examples/pull/75/commits/642eac5dfcc1acad856cf800917446e12cf3cddd)

gabi-250:
The `SIGPROT` issue is fixed in [0d7ed73](https://github.com/capablevms/cheri-examples/pull/75/commits/0d7ed73b030f18758efeb51aaaeea7fa0268c6c9). The solution is to use `adr` instead of `ldr` to load the address of the selected clean function into a register (TIL `ldr x13, =label` loads the address of `label` from a literal pool. The indirect load caused a CHERI protection violation because the literal pool was past `switch_compartment_end`).

gabi-250:
The code now works as expected, so it's almost ready for review (it achieves what I described in [this comment](https://github.com/capablevms/cheri-examples/pull/75#discussion_r1011633053)).

Still to-do:
- [ ] `switch_compartment` takes an additional argument in `x13`, and completely ignores the calling convention. Before merging, I want to update `switch_compartment` to follow the calling convention
- [ ] the `base` example still has its own Makefile (ideally it should be able to use the top-level one like the other examples)
- [ ] `switch_compartment.S` should probably be relocated to `common/` subdir or something

gabi-250:
> The code now works as expected, so it's almost ready for review (it achieves what I described in [this comment](https://github.com/capablevms/cheri-examples/pull/75#discussion_r1011633053)).
> 
> Still to-do:
> 
>     * [ ]  `switch_compartment` takes an additional argument in `x13`, and completely ignores the calling convention. Before merging, I want to update `switch_compartment` to follow the calling convention
>     * [ ]  the `base` example still has its own Makefile (ideally it should be able to use the top-level one like the other examples)
>     * [ ]  `switch_compartment.S` should probably be relocated to `common/` subdir or something


As discussed offline, these improvements can be done in separate PRs (in the interest of keeping this one small and easy to review).

gabi-250:
I reverted [52e92e5](https://github.com/capablevms/cheri-examples/pull/75/commits/52e92e513db4d282cc399103626e0305787d88cc) a) because it probably belongs in a separate PR and b) it doesn't quite work :grinning:.

If you wish, I can squash some of these commits to make the whole thing easier to review.

hybrid/compartment_examples/inter_comp_call/secure/shared.S
```diff
@@ -20,10 +18,8 @@
 executive_switch:
     mov      c29, c0
     mov      x0, #0
+    // Clean up x20 after returning from compartment switch
+    mov      x13, #0
```
gabi-250, line 22: This is the `leak_x20` argument of `switch_compartment`. I don't like this approach as much as the previous one (where we include the appropriate `clean` implementation based on a flag) because I don't think the compartments should have the power to select a `clean` implementation.

hybrid/compartment_examples/inter_comp_call/secure/compartments.s
```diff
@@ -18,14 +18,14 @@ comp_f_fn:
     // `switch_compartment`.
     str       clr, [sp, #-16]!
 
-    // Retrieve local capability containing switcher information for `ldpblr`
+    // Retrieve local capability containing switcher information for `pdlblr`
```
0152la, line 21: Was this always `pdlblr`? If so, it should be `ldpblr`.

hybrid/compartment_examples/inter_comp_call/secure/shared.S
0152la: I agree that it makes the point that the issue can be more subtle and outside the compartments' hands, as we can easily imagine the compartment being user code, and a the transition code being either in a library, or perhaps even the kernel.

hybrid/compartment_examples/inter_comp_call/secure_switch_compartment.S
```diff
@@ -7,45 +7,44 @@
 
 /*
  * The compartment switch function. Expects compartment information to be
- * stored in memory (defined by the capability stored in register `c1`).
+ * stored in memory (defined by the capability stored in register `c29`).
```
0152la, line 10: I see all these changes away from `c29`, which as discussed, matters for the purposes of `ldpblr`. If there is no  top-level comment mentioning how this is a special register (again, for the purposes for `ldpblr`), perhaps one is in order.

hybrid/compartment_examples/inter_comp_call/build/Makefile.inter_comp_call
```diff
@@ -31,6 +31,12 @@ clean:
 		rmdir bin --ignore-fail-on-non-empty; \
 	fi
 
+.SECONDEXPANSION:
```
0152la, line 34: I'm not too familiar with advanced makefile usage. Is this `.SECONDEXPANSION` getting called implicitly somewhere? Additionally, perhaps it would be useful for me if you show me the sequence of commands to compile/execute one of these examples with this new infrastructure, and how the artifacts look like.

hybrid/compartment_examples/inter_comp_call/secure/compartments.s
gabi-250: Oops, I thought I'd fixed it (on master this is only misspelled as `pdlblr` in `secure` and `secure-update_ddc`)

0152la:
> If you wish, I can squash some of these commits to make the whole thing easier to review.

This PR is indeed getting a bit out of hand. Let's add commits addressing current comment chains, and then decide where to go from there. Personally, I don't understand what happened with https://github.com/capablevms/cheri-examples/pull/75/commits/8f67504a0231e747a113fd3769da81add68329fe, as it seems a lot of files just upped and moved, while the original commit to be reverted had a single file moved.

hybrid/compartment_examples/inter_comp_call/secure_switch_compartment.S
gabi-250: Yeah, I didn't mean to push [52e92e5](https://github.com/capablevms/cheri-examples/pull/75/commits/52e92e513db4d282cc399103626e0305787d88cc) (I reverted it in [8f67504](https://github.com/capablevms/cheri-examples/pull/75/commits/8f67504a0231e747a113fd3769da81add68329fe)). I can squash all of this away to make things easier to review.

gabi-250:
If I squash it'll be a bit clearer what this PR intends to achieve (that one commit I didn't mean to push and that I later reverted doesn't help either!).

gabi-250:
I only want to squash [52e92e5](https://github.com/capablevms/cheri-examples/commit/52e92e513db4d282cc399103626e0305787d88cc) and the commit that reverts it. Is that alright?

gabi-250:
Actually, let me just reset HEAD to 28a439cd159bf6c0b67035f133792fa5a4dec0ae and cherry-pick aaac557e6dbb8ce535d69cbac706d611bad287ec on top (to minimize the number of rewritten commits). Is that cool with you?

0152la:
The history [1] of the file touched by https://github.com/capablevms/cheri-examples/commit/aaac557e6dbb8ce535d69cbac706d611bad287ec suggests it was affected by the two commits that would be deleted by resetting to https://github.com/capablevms/cheri-examples/commit/28a439cd159bf6c0b67035f133792fa5a4dec0ae. As such, please just reset, and push a new commit, rather than cherry-picking the existing one.

[1] https://github.com/capablevms/cheri-examples/commits/aaac557e6dbb8ce535d69cbac706d611bad287ec/hybrid/compartment_examples/inter_comp_call/build/Makefile.inter_comp_call

gabi-250:
> As such, please just reset, and push a new commit, rather than cherry-picking the existing one.

What do you mean? Cherry-picking will create a new commit (with a new SHA) with the same contents.

0152la:
Ok; just do as you think fit and we can see the final result.

gabi-250:
Done

0152la:
Alright. We should take care of the existing comment chains appropriately; if some of them are not applicable anymore due to the history change, please indicate as such.

hybrid/compartment_examples/inter_comp_call/clean_leak_ddc.s
gabi-250: Addressed in 2edd256406a0480c26bb8051a33c7ba53fdc755a

hybrid/compartment_examples/inter_comp_call/switch_compartment.S
gabi-250: I pushed the other implementation option in 2edd256406a0480c26bb8051a33c7ba53fdc755a

hybrid/compartment_examples/inter_comp_call/secure-update_ddc/shared-update_ddc.S
gabi-250: As of 2edd256406a0480c26bb8051a33c7ba53fdc755a, this no longer exists (however, if we decide to go with the original approach, I can go back and document it)

hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/Makefile.morello-hybrid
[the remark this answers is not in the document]
gabi-250: > Do the other makefiles in the other example folders not need any similar changes?

Not all of them required this change - some actually `#include`d `switch_compartment.S` from `shared.S`, so there was no need to assemble and link `switch_compartment.S`.

> And maybe further, I think moving these makefiles out of each individual example folder would be a good goal, although I'm sure I must've tried this and given up for some reason or another...

Addressed in d1e3a2598152c969ac4acbf45ee23b96ff7ef42f and 2edd256406a0480c26bb8051a33c7ba53fdc755a (I can squash them into a single commit if it makes things easier to review).

hybrid/compartment_examples/inter_comp_call/secure_switch_compartment.S
gabi-250: Not applicable as this change was reverted.

hybrid/compartment_examples/inter_comp_call/secure/compartments.s
gabi-250: Fixed in c4d34fed27a2dfba8bddb2dfbd9060dcce53d379

gabi-250:
> This PR is indeed getting a bit out of hand. Let's add commits addressing unresolved comment chains, and then decide where to go from there. Personally, I don't understand what happened with [8f67504](https://github.com/capablevms/cheri-examples/commit/8f67504a0231e747a113fd3769da81add68329fe), as it seems a lot of files just upped and moved, while the original commit to be reverted had a single file moved.

The original commit (52e92e513db4d282cc399103626e0305787d88cc) also moved a lot of files - the revert commit just reverts all those changes. Either way, I've removed both as discussed as they're outside the scope of this PR.

ltratt:
I think this is now awaiting a review from @0152la, but I might have got confused, and there should be another commit two?

0152la:
There are some changes I haven't reviewed yet, which I'll go over shortly; unsure if more commits are expected at this stage, but there was an issue with a commit that wasn't meant to be pushed this PR and we decided to revert.

hybrid/compartment_examples/inter_comp_call/switch_compartment.S
0152la: Since https://github.com/capablevms/cheri-examples/commit/2edd256406a0480c26bb8051a33c7ba53fdc755a is a fairly large commit, can you be more specific where I have to look for these related changes?

hybrid/compartment_examples/inter_comp_call/switch_compartment.S
```diff
@@ -41,14 +46,29 @@ switch_compartment:
     // Install compartment DDC
     msr       DDC, c1
 
-    // Save old DDC (c2), old SP (x12), old LR (clr) on stack
-    stp       c2, clr, [sp, #-48]!
+    // Save old DDC (c2),  old LR (clr), old SP (x12) on stack
+    stp       c2, clr, [sp, #-64]!
```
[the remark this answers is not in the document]
0152la: Ok, I see. Later on, you save the `clean` target on the stack as well. I would do all the stack saving together, to make this obvious. I am unsure if we need to put this on the stack though. I also dislike how the clean `bl clean+4` changed to something ugly. I need to see if we can do something like `blr x13+4` if we want to keep this approach.

hybrid/compartment_examples/inter_comp_call/build/Makefile.inter_comp_call
0152la: Ok, let me rephrase. One question was "is `.SECONDEXPANSION` a `make` implicit" thing, which you said yes (complete with link). The other question I'll rephrase as "does this change anything in how we call `make`, or in what it produces"?

gabi-250: >  The other question I'll rephrase as "does this change anything in how we call make, or in what it produces"?

`so you don't need to change your make invocation for it to happen` - in other words, it doesn't change anything in how we call make and it doesn't change any of the resulting artifacts.

gabi-250:
> There are some changes I haven't reviewed yet, which I'll go over shortly; unsure if more commits are expected at this stage, 

Sorry, I should've been clearer: I have no additional commits to push. I think we need to discuss [this comment](https://github.com/capablevms/cheri-examples/pull/75#discussion_r1013868218) before I start addressing any of the new comments. TLDR I think the current approach is fundamentally flawed (I happen to prefer the code duplication over it, because I don't think compartments should be able to tell `switch_compartment` which `call` implementation to use). From my point of view, we should either
* switch HEAD back to https://github.com/capablevms/cheri-examples/pull/75/commits/5f94374246fa284ebef72a7ad049909e20fbe55a (as this version reduces code duplication without giving compartments control over which `clean` implementation is called
* decline this PR altogether

I'm happy with either outcome.

gabi-250:
I'm going to decline this PR as it's become rather bloated and hard to review. I'm going to open a number of smaller PR for the minor improvements (typo fixes) we identified in this one, as well as for the refactoring.

PR #76, opened 2022-11-07
Refactor `switch_compartment.s` v2

gabi-250:
### Context

As the diffs below show, the `switch_compartment.s`s of the 3 `secure` examples are almost identical:
```
gabi@areca ~/c/cheri-examples > diff hybrid/compartment_examples/inter_comp_call/secure/switch_compartment.s hybrid/compartment_examples/inter_comp_call/secure-try_deref/switch_compartment.s
38c38
<     // Save old DDC (c2), old SP (x12), old CLR (clr) on stack
---
>     // Save old DDC (c2), old SP (x12), old LR (clr) on stack
gabi@areca ~/c/cheri-examples > diff hybrid/compartment_examples/inter_comp_call/secure/switch_compartment.s hybrid/compartment_examples/inter_comp_call/secure-redirect_clr/switch_compartment.S
5a6,8
>
> #include "../include/secure.h"
>
6a10,11
> .global switch_compartment_end
>
```

Since the only real difference between them is the exported `switch_compartment_end` symbol, it's safe to refactor the 3 examples to use the same `switch_compartment.s`. Deduplicating the code makes it easier to understand the differences between the `secure_*` variants.

### Implementation

This change refactors the compartment switcher (`switch_compartment.S`) used by the `secure*` examples.

Some `inter_comp_call` examples use a correct `clean` implementation, whereas the `secure-update_ddc` example uses one that intentionally fails to clear `x20`. This commit introduces the `INTER_COMP_LEAK_X20` flag, which can be used to  include one of the two implementations in `switch_compartment.S` to further deduplicate the code. `INTER_COMP_LEAK_X20` is only needed for the `secure-update_ddc` example, which relies on a "broken" `clean` implementation to validate certain security properties. `INTER_COMP_LEAK_X20` is set in the top-level makefile of the `secure` examples.

This change also restructures the `inter_comp_call` directory:
```
.
├── base
│   ├── include
│   │   └── base.h
│   ├── main.c
│   ├── Makefile.morello-hybrid
│   ├── README.md
│   └── shared.S
└── secure
    ├── build
    │   └── Makefile.inter_comp_call
    ├── include
    │   └── secure.h
    ├── Makefile.morello-hybrid
    ├── README.md
    ├── secure
    │   ├── compartments.s
    │   ├── main.c
    │   └── shared.S
    ├── secure-redirect_clr
    │   ├── compartments-redirect_clr.S
    │   ├── main.c
    │   └── shared.S
    ├── secure-try_deref
    │   ├── compartments-try_deref.s
    │   ├── main.c
    │   └── shared.S
    ├── secure-update_ddc
    │   ├── compartments-update_ddc.S
    │   ├── main.c
    │   └── shared-update_ddc.S
    └── shared
        ├── include
        └── switch_compartment.S
```

**Note**: the refactoring of the `base` inter_comp_call example is outside the scope of this change. The `switch_compartment.S` variant from `base` is different from the `switch_compartment.S` version used for the `secure*` examples, so much so that its refactoring warrants a separate PR.


ltratt:
Do we document that the `secure-update_ddc` example leaks? [Basically I don't want someone to come along later and think that a possibly subtly broken bit of code is actually correct!]

gabi-250:
> Do we document that the `secure-update_ddc` example leaks? [Basically I don't want someone to come along later and think that a possibly subtly broken bit of code is actually correct!]

It's documented in the `secure-update_ddc/main.c`:
```
/* This example attempts to modify the DDC of `switch_compartment`, assuming it
 * knows the address where it resides (leaked via `c20` in `executive_switch()`
 * in `shared-update_ddc.S`, and excluded from the `clean()` call in
 * `switch_compartment()). However, as it lies outside the DDC of the
 * compartment, attempting to load from the respective memory address causes a
 * fault (as attempted in `comp_f_fn` of `compartments-update_ddc.S`).
 */
 ```

File: hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call [hunk 1]
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2020-2021 The CapableVMs "CHERI Examples" Contributors.
```
0152la, line 1: Year

ltratt:
I might not have quite understood one or the other comment. What I thought you said was "the example leaks data" (with the implication that it carries on executing) but the comment suggests "the example tries to leak but crashes when it does so"?

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
```diff
@@ -0,0 +1,57 @@
...
+$(error build/Makefile.vars should set BINDIR)
+endif
+
+ifdef RUNDIR
+RUNDIR := $(RUNDIR)/inter_comp_call
+else
+RUNDIR := inter_comp_call
+endif
+
+UPDATE_DDC_EXAMPLE := secure-update_ddc
```
0152la, line 17: Is this a list, or how would it look like adding more examples to this variable? I would perhaps say a few words what this will do, even if it is mentioned a bit later on.

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
```diff
@@ -0,0 +1,57 @@
...
+$(BINDIR)/%: $$(wildcard %/*.s) $$(wildcard %/*.S) $$(wildcard %/*.c) $(SHARED_SOURCES)
+	@mkdir -p $(BINDIR)
+	$(CC) $(CFLAGS) -L $(BINDIR) $+ -o $@
+
+# switch_compartment.S conditionally includes one of the two clean
+# implementations. Setting INTER_COMP_LEAK_X20 tells switch-compartment.S to
+# select the "broken" version of clean (which simulates leaking a register).
+# In the UPDATE_DDC_EXAMPLE example, x20 contains the DDC of the compartment
+# switcher.
+run-inter_comp_call-$(UPDATE_DDC_EXAMPLE): CFLAGS += -D INTER_COMP_LEAK_X20
```
0152la, line 47: Let me try to understand: in `UPDATE_DDC_EXAMPLE`, we have a list of examples for which we want to run with the "leaky" version of `clean`. This basically says "if you do `make run-inter_comp_call-{example}`, if that `{example}` is one of `UPDATE_DDC_EXAMPLE`, we will be adding the "leaky flag"". Did I get that right?

build/Makefile.simple
```diff
@@ -31,6 +31,9 @@ clean:
 		rmdir bin --ignore-fail-on-non-empty; \
 	fi
 
+%.s: %.S
```
0152la, line 34: Can you tell me what this does? And presumably this doesn't affect any of the other examples.

hybrid/compartment_examples/inter_comp_call/base/Makefile.morello-hybrid
```diff
@@ -3,7 +3,8 @@
 
 SHARED_SOURCES := shared.s
 CFILES := $(wildcard *.c)
-ROOTDIR=../../../..
+ROOTDIR = ../../../..
+CFLAGS += -I$(ROOTDIR)/include -I$(ROOTDIR)/hybrid/include -Iinclude
```
0152la, line 7: Do we need all these three includes?

hybrid/compartment_examples/inter_comp_call/base/README.md
```diff
@@ -0,0 +1,10 @@
+# "inter_comp_call-base" example
```
0152la, line 1: Wouldn't this README be better placed in the root `inter_comp_call` folder, rather than `base`?

hybrid/compartment_examples/inter_comp_call/secure/README.md
```diff
@@ -0,0 +1,10 @@
+# "inter_comp_call-secure*" examples
```
0152la, line 1: I see we have two READMEs now. What is the intent behind there being two?

gabi-250:
(note: I'm describing the way things were before the refactoring)

This is a bit confusing because "leak" is an overloaded term (in this example, we're not leaking data, we're actually just "leaking" a register across compartment switches).

`switch_compartment` has a [`clean`](https://github.com/capablevms/cheri-examples/blob/066c5c7d0d3edb18e284e228ea7c982ec458ab2b/hybrid/compartment_examples/inter_comp_call/secure/switch_compartment.s#L75-L123) function, which is used to clear most of the registers before/after a compartment switch.

The `secure-update_ddc` example has a separate copy of `switch_compartment`, which uses an intentionally broken `clean` implementation. It's broken in the sense that it ["leaks" (i.e. fails to clear) the `x20`/`c20` register](https://github.com/capablevms/cheri-examples/blob/066c5c7d0d3edb18e284e228ea7c982ec458ab2b/hybrid/compartment_examples/inter_comp_call/secure-update_ddc/switch_compartment-update_ddc.S#L116). The malicious compartment then attempts to [use](https://github.com/capablevms/cheri-examples/blob/066c5c7d0d3edb18e284e228ea7c982ec458ab2b/hybrid/compartment_examples/inter_comp_call/secure-update_ddc/compartments-update_ddc.S#L23-L28) the leaked `x20` register (which results in a crash).

0152la (commented):
I still think this PR tries to do too many things at once. Particularly, I am confused about the changes to `comp_setup/shared.S`. I don't see any mention in the PR body about any changes that should affect that example.

hybrid/compartment_examples/inter_comp_call/secure/secure/compartments.s
```diff
@@ -3,6 +3,11 @@
  * appropriately to cover `switch_compartment`.
  */
 .type comp_f_fn, "function"
+.globl comp_f_fn
```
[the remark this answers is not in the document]
gabi-250: They're equivalent (I type it as `.globl` out of habit).

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
gabi-250: It's meant to be just one (if we want this to be a list instead, I'll need to rework the part where `UPDATE_DDC_EXAMPLE` is used).

0152la:
> Do we document that the `secure-update_ddc` example leaks? [Basically I don't want someone to come along later and think that a possibly subtly broken bit of code is actually correct!]

We wanted to have some counter-examples, which would (a) strengthen the belief that the "secure" implementation is indeed secure (which we know it's not), and (b) ensure that CHERI can capture those counter-examples. There should be indication in the top comment of each test whether it fails or not, and there should also be a list in the testing shell script, which says which should succeed and which should fail.

It is a bit odd that under the proposed `secure` folder, we'll have subfolders showing counter-examples.

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
gabi-250: Yep you got it right, except `UPDATE_DDC_EXAMPLE` can't be a list (this assumes `UPDATE_DDC_EXAMPLE` is a single example).

build/Makefile.simple
gabi-250: It overrides the default recipe for compiling `.S` files into `.s` files. The default recipe runs the preprocessor on an `.S` file to obtain the corresponding `.s`, but it completely ignores the `CFLAGS`. I want it to use the `CFLAGS` (to which I appended the relevant include paths), because that allows us to remove the relative paths from any `#include`s we might have in our `.S` files (it allows us to go from `#include ../../../../../include/common.h` to `#include common.h`)

ltratt:
@gabi-250 

> The secure-update_ddc example has a separate copy of switch_compartment, which uses an intentionally broken clean implementation. It's broken in the sense that it ["leaks" (i.e. fails to clear) the x20/c20 register](https://github.com/capablevms/cheri-examples/blob/066c5c7d0d3edb18e284e228ea7c982ec458ab2b/hybrid/compartment_examples/inter_comp_call/secure-update_ddc/switch_compartment-update_ddc.S#L116). The malicious compartment then attempts to [use](https://github.com/capablevms/cheri-examples/blob/066c5c7d0d3edb18e284e228ea7c982ec458ab2b/hybrid/compartment_examples/inter_comp_call/secure-update_ddc/compartments-update_ddc.S#L23-L28) the leaked x20 register (which results in a crash).

Got it. Judging by @0152la's comment, we can perhaps find a better name for this, but I get what the intention is now.

hybrid/compartment_examples/inter_comp_call/secure/README.md
gabi-250: They're separate because you are expected to run the `secure` examples from the `hybrid/compartment_examples/inter_comp_call/secure` directory and the `base` example from `hybrid/compartment_examples/inter_comp_call/base` directory (the READMEs are in the two directory you're supposed the examples from). We can merge them, but the instructions would then need to specify that you need to `cd secure` or `cd base` before running `make`.

gabi-250:
> It is a bit odd that under the proposed secure folder, we'll have subfolders showing counter-examples.

Good point, maybe I can rename it to something else (or rethink the structure a bit if you prefer). The only reason I grouped them that way is because the `shared` directory where I moved `switch_compartment.S` and the two clean implementation only pertains to the `secure*` examples - the `base` example still has its own version of `switch_compartment.S`.

hybrid/compartment_examples/inter_comp_call/base/README.md
gabi-250: Yep, I can merge the two READMEs.

build/Makefile.simple
gabi-250: It shouldn't affect any of the other examples (the recipe is the mostly same as before, the only difference being that it uses the `CFLAGS` we define)

gabi-250:
> Particularly, I am confused about the changes to comp_setup/shared.S. I don't see any mention in the PR body about any changes that should affect that example.

Oops, the change from `comp_setup/shared.S` is a find/replace gone wrong.

gabi-250:
> > Particularly, I am confused about the changes to comp_setup/shared.S. I don't see any mention in the PR body about any changes that should affect that example.
> 
> Oops, the change from `comp_setup/shared.S` is a find/replace gone wrong.

Fixed in https://github.com/capablevms/cheri-examples/pull/76/commits/8781c0012d44b3df73587df900121b8640484bfd

hybrid/compartment_examples/inter_comp_call/secure/secure/compartments.s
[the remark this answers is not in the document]
gabi-250: I switched to `.global` in https://github.com/capablevms/cheri-examples/pull/76/commits/23843348d1847c0d1e724ff4f2b3a3beb85d36f2

hybrid/compartment_examples/inter_comp_call/secure/secure/compartments.s
```diff
@@ -3,6 +3,11 @@
...
+.globl comp_f_fn_end
+.globl comp_g_fn
```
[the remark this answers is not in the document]
gabi-250: Fixed in https://github.com/capablevms/cheri-examples/pull/76/commits/ab2d493409577f739a58f7808b59685534d533a2

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
0152la: I think if it's dynamic it should either be a list, so it can be updated at will, or it should be static, so we hardcode saying which examples we want to leak. The cleanest and most future proof would be if we can somehow make it a list, but I don't know how difficult that might be. An alternative would be adding a new `make` command, like `make run-leak-{example}` or something like that.

hybrid/compartment_examples/inter_comp_call/base/Makefile.morello-hybrid
0152la: This is a concern since we have so many include files lying around, but that's probably better saved for a future, larger scale refactoring.

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
gabi-250: I think this can easily be rewritten to support lists. I'll give it a try and push a commit for you to review.

hybrid/compartment_examples/inter_comp_call/secure/README.md
0152la: Right, I see. I think that would be better, to have one, and specify the two example sets (`base` and `secure`), but I have no strong feelings either way.

hybrid/compartment_examples/inter_comp_call/base/README.md
0152la: Refer to https://github.com/capablevms/cheri-examples/pull/76#discussion_r1015610691.

0152la:
> I can't decouple most of these changes (to refactor the examples to use the same `Makefile`, various other changes were needed). I think the only thing I _can_ extract into a separate PR is the change for specifying include paths (needed to save us from having to type long relative paths when including headers).

With the `comp_setup` reversal, I think it's better. The eventual commit should include a mention that we also refactor the build infrastructure.

hybrid/compartment_examples/inter_comp_call/base/Makefile.morello-hybrid
gabi-250: I switched back to relative includes in https://github.com/capablevms/cheri-examples/pull/76/commits/4ffda3fb3ba60d245ca3144a7f5b9a5e72485ac8

hybrid/compartment_examples/inter_comp_call/secure/README.md
gabi-250: I merged the READMEs into a single top-level README: https://github.com/capablevms/cheri-examples/pull/76/commits/7fb7dac0311712fe0b17a8477c298912f7292051

gabi-250:
I renamed `secure` to `malicious_compartments`: in https://github.com/capablevms/cheri-examples/pull/76/commits/e164582aff5a445ebd570b16a5cec0947fc6b520 

(`malicious_compartments` isn't a great name either, but it's more fitting than `secure` because all of those examples have compartments that misbehave in one way or another)

hybrid/compartment_examples/inter_comp_call/README.md
```diff
@@ -0,0 +1,18 @@
...
+$ cd base && make -f Makefile.morello-hybrid run-main
+```
+
+* to run one of the `secure` examples:
+
+```
+$ cd secure && make -f Makefile.morello-hybrid run-inter_comp_call-<example>
+```
+
+Refer to the top-level [README](../../../../README.md) for usage details.
```
0152la, line 22: I think there's one too many `..`.

hybrid/compartment_examples/inter_comp_call/secure/build/Makefile.inter_comp_call
0152la: Think it looks cleaner as well. Nice.

hybrid/compartment_examples/inter_comp_call/README.md
gabi-250: Good catch. Fixed in https://github.com/capablevms/cheri-examples/pull/76/commits/faa109b36b3f4a4245e8cb88c73585b2b10bbabe

0152la:
Please squash.

gabi-250:
Squashed and amended the commit message to also mention the build system change.

gabi-250:
I pushed another commit to fix the tests (it'll need to be squashed)

gabi-250:
bors try

0152la:
Please squash.

gabi-250:
squashed

PR #77, opened 2022-11-11
Replace unreachable `ret` instructions with `udf`.

gabi-250:
This replaces several unreachable `ret` instructions with `udf` (the `permanently undefined` instruction). Using `udf` instead of `ret` makes it more obvious that `switch_compartment` doesn't return to `executive_switch` (it returns to the caller of `executive_switch`).

### DIFF
--- a/hybrid/compartment_examples/inter_comp_call/malicious_compartments/secure-redirect_clr/shared.S
+++ b/hybrid/compartment_examples/inter_comp_call/malicious_compartments/secure-redirect_clr/shared.S
@@ -11,4 +11,5 @@ executive_switch:
     mov      x0, #0
     cvtp     clr, lr
     b        switch_compartment
-    ret      clr
+    // Unreachable (switch_compartment returns to main)
+    udf      #0xdead
--- a/hybrid/compartment_examples/inter_comp_call/malicious_compartments/secure-try_deref/shared.S
+++ b/hybrid/compartment_examples/inter_comp_call/malicious_compartments/secure-try_deref/shared.S
@@ -11,4 +11,5 @@ executive_switch:
     mov      x0, #0
     cvtp     clr, lr
     b        switch_compartment
-    ret      clr
+    // Unreachable (switch_compartment returns to main)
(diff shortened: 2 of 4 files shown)


hybrid/compartment_examples/inter_comp_call/malicious_compartments/secure-redirect_clr/shared.S
```diff
@@ -11,4 +11,5 @@ executive_switch:
     mov      x0, #0
     cvtp     clr, lr
```
gabi-250, line 12: @0152la As far as I can tell, this instruction can be removed too at this point (I don't think we need `lr` converted to a capability).

0152la: I *think* this is needed to revert the capability metadata set in `switch_compartment` itself [1]. However, without a `gdb` in front of me going instruction by instruction, I can't say certainly that this cannot be remove. The program *should* crash after returning from `switch_compartment` [2] if we do not explicitly provide a capability. Also should add if this `cvtp` is removed, the corresponding return in `switch_compartment` [2] needs to be changed to a simple `ret`.

[1] https://github.com/gabi-250/cheri-examples/blob/bd4d47f71537402261bd4b4bd007322ab0c0e71a/hybrid/compartment_examples/inter_comp_call/malicious_compartments/shared/switch_compartment.S#L66
[2] https://github.com/gabi-250/cheri-examples/blob/bd4d47f71537402261bd4b4bd007322ab0c0e71a/hybrid/compartment_examples/inter_comp_call/malicious_compartments/shared/switch_compartment.S#L79

0152la: Alright, so can I read that as we both agree the `cvtp` is needed then?

gabi-250: yep!

PR #78, opened 2023-01-20
Added new example compare_platforms

tel3path:
This example can be run in a CHERI or non-CHERI environment. 

In a CHERI environment, it uses capabilities to set a struct to read-only, tries to write to it, and crashes. 

In a non-CHERI environment it uses a member of the struct to set permissions. It overflows the bounds of adjacent objects to add write permissions to the read-only struct, and then changes another member of the struct.


ltratt:
This ended up as a submodule, rather than a directory with files in. Can you force push a "normal" commit please?

tel3path:
Done

compare_platforms/include/review.h
```diff
@@ -0,0 +1,151 @@
...
+    char *publicreview;
+    char *permissions;
+};
+
+/**
+ * Check this review's permissions.
+ * param: review, the struct with the details of the review whose permissions we are checking
+ * return: true if the review has write permissions, false otherwise.
+**/
+bool can_write(struct review *rv) {
```
ltratt, line 37: I think this function can be just `return strchr(rv->permissions, "w") == NULL;` ?

ltratt:
In general, in C, we don't want to put code in `.h` files (it causes all sorts of compilation wonkery; this isn't necessarily true for C++, however), so we should just move the contents of the `.h` file into the `.c` file.

ltratt:
I think I've got the general idea, but there's 300 lines of code, and probably the core part of the example is about 40-50LoC, perhaps 70LoC with the non-CHERI bits. Can we cut out some of the extraneous bits so that it's easier to concentrate on the important parts?

compare_platforms/build/Makefile.review
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2023 The CapableVMs "CHERI Examples" Contributors.
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+# Common Makefile for the privesc examples.
```
0152la, line 4: I believe this is not the "privesc" example. 

I'm unsure what purpose this Makefile has. I'm used to the convention of having "Makefile.<platform>" to build for target <platform>, but this seems to refer to a header file?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+int main()
+{
+    char *username, *realname, *privatereview, *publicreview, *permissions;
```
0152la, line 35: More out of curiosity, but any reason we define all of these here on top, rather than inline?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+    
+    // Our new review, which will have read only permissions.
+    struct review review;
+    
+    size_t smallsz = 0x20;
+    size_t biggersz = 0x1001;
```
0152la, line 41: What do these two values represent?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+    
+    // Populate the members of the review struct.
+    username = malloc(smallsz);
+    strcpy(username, "Reviewer 2");    
+    size_t szn = actual_input_length(username, smallsz);
```
0152la, line 46: I'm unsure of what's happening with `actual_input_length`. Is it here to provide the additional size checks? If so, maybe a more appropriate name is in order, or could move the whole allocate, initialize, check process to a function, to reduce duplication.

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+    
+        print_details(&review);
+        
+        b_improved = try_improve_publicreview(&review, biggersz, false);
+        
+        print_details(&review);
+    }
+
+    return 0;
+}
```
0152la, line 145: Are the `malloc`s freed somewhere?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+    
+    review.username = username;
+    review.realname = realname;
+    review.privatereview = privatereview;
+    review.publicreview = publicreview;
+    review.permissions = permissions;
+    print_details(&review);
+
+#ifdef __CHERI_PURE_CAPABILITY__
+    printf("\nsmallsz=%zx, CRRL(smallsz)=%zx\n", smallsz,
```
0152la, line 76: What's `CRRL` here?

compare_platforms/include/review.h
```diff
@@ -0,0 +1,151 @@
...
+ * param: reqsz, a size_t representing the amount of memory to allocate for the new review.
+ * param: b_weak, a boolean that is true if we are weakly enforcing (i.e. using can_write) the 
+ * read-only permissions, false otherwise.
+ * return: false if review is read-only, true if review is writable.
+**/
+bool try_improve_publicreview(struct review *rv, size_t reqsz, bool b_weak) {
+    bool b_writable = false;
+    
+    char *newpublicreview = malloc(reqsz);
+    strcpy(newpublicreview, "5 ***** strong accept. This author deserves two Nobels and an ice cream.\n");
```
0152la, line 87: I have a feeling the string is better off as an argument to the function.

compare_platforms/include/review.h
```diff
@@ -0,0 +1,151 @@
...
+
+
+
+#ifdef __CHERI_PURE_CAPABILITY__
+/**
+ * Set all the struct members to read-only.
+ * param: rv, the struct representing the review to display
+ * return: a review with permissions restricted to read-only
+ **/
+struct review *set_read_only(struct review *rv)
```
0152la, line 142: Hmm, this function being only defined in a CHERI environment might lead to issues during compilation. I would suggest we move the `ifdef` inside the function, and if we are not in a CHERI environment, we `assert(false)`.

compare_platforms/include/review.h
```diff
@@ -0,0 +1,151 @@
...
+    strcpy(newpublicreview, "5 ***** strong accept. This author deserves two Nobels and an ice cream.\n");
+
+    printf("\nAttempting to change the public review.\n");
+    #ifdef __CHERI_PURE_CAPABILITY__
+        if((can_write(rv) == false) && (b_weak == true)) {
+            printf("You do not have permission to change the public review.\n");
+            return false;
+        } else {
+            // if the review is set to read-only, this will crash.
+            b_writable = change_publicreview(rv, newpublicreview);
```
0152la, line 96: I think this `ifdef` here needs re-organizing. There is one in `change_publicreview` as well. Could we reduce the number of places we check for CHERI minimally, so we can say "aha, here's the exact place where divergence happens"?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+        print_details(ro_review);
+    }
+# else
+    // Try editing the public review in a non-CHERI environment,
+    // by simply checking the write permissions and refusing if none are found.
+    b_improved = try_improve_publicreview(&review, biggersz, false);
+#endif
+
+    // This code will be unreachable in a CHERI environment,
+    // but it would crash whether or not the struct had been set to read-only.
```
0152la, line 123: I am assuming it's unreachable due to the read-only aspect. Could we have a switch to turn that off, so that we can see the crash in CHERI as well, when we set the switch?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+    if(b_improved == false) {
+        printf("\nOverflowing reviewer realname by 1\n");
```
0152la, line 125: I might be talking gibberish, but I would be careful with overflowing by just 1, due to the "one past the end" rule. I'm unsure if it's applicable here, but just in case.

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,147 @@
...
+    
+        // Overflow the realname even more, which in a CHERI
+        // environment should crash if it hasn't already.
+        const size_t oversz = review.permissions - review.realname + 2 - smallsz;
+        printf("\nOverflowing reviewer realname by %zx\n", oversz);
+        memset(review.realname + smallsz+2, 'w', oversz);
+    
+        print_details(&review);
+        
+        b_improved = try_improve_publicreview(&review, biggersz, false);
```
0152la, line 139: Is there any difference between this and the call above?

0152la (commented):
I think the main weakness is a lot of `ifdef __CHERI__`, intertwined in various function calls. I'd like to see if we can clean those up a bit, to be more clear where the divergence happens.

There are also two files which seem they might be extraneous: `patch` and `.DS_Store`.

compare_platforms/compare_platforms_read_only.c
tel3path: I've started getting an exec format error when I try to run this in morello-purecap. This started after I added a switch, but reverting hasn't fixed it.

compare_platforms/include/review.h
tel3path: On lines 46-48, can_write has been reduced to:

return (strchr(rv->permissions, 'w') != NULL);

tel3path:
> 



> In general, in C, we don't want to put code in `.h` files (it causes all sorts of compilation wonkery; this isn't necessarily true for C++, however), so we should just move the contents of the `.h` file into the `.c` file.

Done.

tel3path:
> I think I've got the general idea, but there's 300 lines of code, and probably the core part of the example is about 40-50LoC, perhaps 70LoC with the non-CHERI bits. Can we cut out some of the extraneous bits so that it's easier to concentrate on the important parts?

Partly done.

compare_platforms/compare_platforms_read_only.c
tel3path: No reason. They are now inline.

compare_platforms/compare_platforms_read_only.c
tel3path: Malloc amounts for the shorter struct members (names) and longer ones (review text).

compare_platforms/compare_platforms_read_only.c
tel3path: It's no longer needed. actual_input_length is now removed.

compare_platforms/compare_platforms_read_only.c
tel3path: "CRoundRepresentableLength(CRRL) accepts a proposed bounds length, and returns a rounded-up size that will be accepted by CSetBoundsExact without throwing an exception." UCAM-CL-TR-951, section 3.5.5.

compare_platforms/include/review.h
tel3path: It is. String is now defined in lines 137-139, and is now an argument to change_publicreview on line 50.

compare_platforms/include/review.h
tel3path: Done.

compare_platforms/compare_platforms_read_only.c
tel3path: Added in lines 148-156.

compare_platforms/compare_platforms_read_only.c
tel3path: I see your point. In the original tutorial, which was for a riscv64-purecap environment, the overflow succeeded on one past the end. In cheri-morello it doesn't. I think this is interesting enough to keep in. I would like to see if it still behaves differently in riscv64-purecap as well, or if it's the same in both environments.

compare_platforms/compare_platforms_read_only.c
tel3path: try_improve_publicreview has been removed and its equivalent functionality folded into change_publicreview, lines 50-61.

tel3path:
Apologies for the gargantuan commit. Will do smaller commits each time from now on.

compare_platforms/compare_platforms_read_only.c
tel3path: Comment header amended to reflect addition of the switch.

compare_platforms/include/review.h
tel3path: Done. CHERI checks are carried out in two places now: inside set_read_only, and in main().

compare_platforms/build/Makefile.review
tel3path: I followed the structure of the original example that I based this one on. As we discussed, it would be better to clean up the makefiles so that all examples can be built with the top-level makefiles. 

For now, I've corrected the comment from "privesc" to "compare_platforms".

tel3path: Untracked and ignored patch and .DS_Store.

compare_platforms/compare_platforms_read_only.c
0152la: Could we make them `const` and add a small comment like "initialization values for struct member sizes"?

compare_platforms/compare_platforms_read_only.c
0152la: Do we expect the CRRL to be the same as the values we set, or not? And if not, will this affect functionality?

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,189 @@
...
+    
+    if(can_write(&review) == false) {
+        printf("\nSetting the review to read-only.\n");
+        struct review *ro_review = set_read_only(&review);
+        assert((cheri_perms_get(ro_review) & CHERI_PERM_STORE) == 0);
+
+        printf("\nThe struct is read-only so trying to change the review will make the program crash.\n");
+        printf("\nTry anyway? y/n: \n");
+        char answer;
+        scanf(" %c", &answer);
```
0152la, line 147: Rather than requesting user input, can we make it a command-line flag?

compare_platforms/compare_platforms_read_only.c
tel3path: With the smaller allocations, we do. With larger allocations, the CRRL might be wider. Length <= 4096 is supposed to have 1-byte alignment. Supposedly this is something for the system components to take care of, not the developer.

compare_platforms/include/review.h
tel3path: > 

Fixed in a8a7797 (a8a7797167ed886112d4223145091bc1b7a54949)

compare_platforms/compare_platforms_read_only.c
0152la: I'm personally unsure if this is needed information to be printed, but I have no strong feelings. If you think it's fine as is, just resolve this thread. Otherwise, change as you see fit and update.

tel3path:
> In general, in C, we don't want to put code in `.h` files (it causes all sorts of compilation wonkery; this isn't necessarily true for C++, however), so we should just move the contents of the `.h` file into the `.c` file.

Fixed in 84d32d4 (84d32d463865d0ff38a349e845ffbed0ea7cd48a)

tel3path:
> This ended up as a submodule, rather than a directory with files in. Can you force push a "normal" commit please?

Fixed in a8a7797167ed886112d4223145091bc1b7a54949

compare_platforms/include/review.h
tel3path: > On lines 46-48, can_write has been reduced to:
> 
> return (strchr(rv->permissions, 'w') != NULL);

Fixed in 84d32d463865d0ff38a349e845ffbed0ea7cd48a

tel3path:
> In general, in C, we don't want to put code in `.h` files (it causes all sorts of compilation wonkery; this isn't necessarily true for C++, however), so we should just move the contents of the `.h` file into the `.c` file.

Fixed in 84d32d463865d0ff38a349e845ffbed0ea7cd48a

compare_platforms/include/review.h
tel3path: > I think this function can be just `return strchr(rv->permissions, "w") == NULL;` ?

Fixed in a41b3e55231a43e96a0da7612f43668dd452f796

compare_platforms/compare_platforms_read_only.c
tel3path: > More out of curiosity, but any reason we define all of these here on top, rather than inline?

Fixed in a41b3e55231a43e96a0da7612f43668dd452f796

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in a41b3e55231a43e96a0da7612f43668dd452f796

> I'm unsure of what's happening with `actual_input_length`. Is it here to provide the additional size checks? If so, maybe a more appropriate name is in order, or could move the whole allocate, initialize, check process to a function, to reduce duplication.

Fixed in a41b3e55231a43e96a0da7612f43668dd452f796

compare_platforms/include/review.h
tel3path: > I have a feeling the string is better off as an argument to the function.

Fixed in a41b3e55231a43e96a0da7612f43668dd452f796

compare_platforms/compare_platforms_read_only.c
tel3path: > Is there any difference between this and the call above?

Fixed in a41b3e55231a43e96a0da7612f43668dd452f796

tel3path:
> I think the main weakness is a lot of `ifdef __CHERI__`, intertwined in various function calls. I'd like to see if we can clean those up a bit, to be more clear where the divergence happens.
> 
Fixed in ccb5c32572367fd59362ebc3b64e557b811a7962

tel3path:
> There are also two files which seem they might be extraneous: `patch` and `.DS_Store`.

Fixed in d616a220f8755304f447b6c90efc8cc2bf027dcb

compare_platforms/include/review.h
tel3path: > I think this `ifdef` here needs re-organizing. There is one in `change_publicreview` as well. Could we reduce the number of places we check for CHERI minimally, so we can say "aha, here's the exact place where divergence happens"?

Fixed in ccb5c32572367fd59362ebc3b64e557b811a7962

compare_platforms/include/review.h
tel3path: > Hmm, this function being only defined in a CHERI environment might lead to issues during compilation. I would suggest we move the `ifdef` inside the function, and if we are not in a CHERI environment, we `assert(false)`.

Fixed in 84d32d463865d0ff38a349e845ffbed0ea7cd48a

compare_platforms/compare_platforms_read_only.c
tel3path: I think it's fine as is.

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in cdc071e3ab3a1248c190f23097a4e8a14dd40167

compare_platforms/build/Makefile.review
tel3path: > I believe this is not the "privesc" example.
> 
Fixed in 84d32d463865d0ff38a349e845ffbed0ea7cd48a

tel3path:
> I think I've got the general idea, but there's 300 lines of code, and probably the core part of the example is about 40-50LoC, perhaps 70LoC with the non-CHERI bits. Can we cut out some of the extraneous bits so that it's easier to concentrate on the important parts?

(Arguably) fixed in ff9f694c6db9bd503f908f2e2dfe35aecf0e7f8b

compare_platforms/compare_platforms_read_only.c
tel3path: They are now. Fixed in 7f3363708787a2d222854123f5e6f213b93c1624

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in 66e958a761fdeb42fa4e944393fd607161ddcedc

compare_platforms/compare_platforms_read_only.c
0152la: Alright; let's keep this comment here to remind ourselves, and once everything else is in order, we can decide how to proceed.

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,202 @@
...
+    printf("Freeing privatereview.\n");
+    free(privatereview);
+    printf("Freeing publicreview.\n");
+    free(publicreview);
+    printf("Freeing permissions.\n");
+    free(permissions);
+
+    return 0;
+}
+
```
0152la, line 194: Lots of whitespace.

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,202 @@
...
+    
+        const size_t oversz = review.permissions - review.realname + 2 - smallsz;
+        overflow_reviewer_realname(smallsz+2, oversz, &review);
+        
+        // If we reached this line, we should have acquired write privileges on the review.
+        b_improved = change_publicreview(&review, newpublicreview, bWeak);
+        print_details(&review);
+    }
+    
+    printf("\nFreeing username.\n");
```
0152la, line 181: I personally think there's too much `printf`ing going on (not only here, but throughout). I don't know if it's worth keeping so much in the merged version. @ltratt thoughts?

tel3path: Fixed in e0adbc70332ea555fd0f0fbdf355c7b7f897f79b

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in d4f4ce7378bf5f83b9edf0406d9e539b31fa17e9

tel3path:
> I think I've got the general idea, but there's 300 lines of code, and probably the core part of the example is about 40-50LoC, perhaps 70LoC with the non-CHERI bits. Can we cut out some of the extraneous bits so that it's easier to concentrate on the important parts?

Fixed in 709c805050cdeed26dadc12a7545d298941f3caa

tel3path:
Further decomposition needed. Fixed in 6b0935d804ebbcd4680632b122517c7df1a56b6f

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,209 @@
...
+    {
+        struct review *ro_review = set_read_only(&review);
+        assert((cheri_perms_get(ro_review) & CHERI_PERM_STORE) == 0);
+
+        if ((argc > 1) && (strcmp(argv[1], "-overwrite") == 0)) 
+        {
+            printf("\nThe struct is read-only so trying to change the review will make the program crash.\n");
+            bWeak = false;
+            b_improved = change_publicreview(ro_review, newpublicreview, bWeak);
+            // This line should be unreachable.
```
0152la, line 187: If a line should be unreachable, it would be best to make it `assert(false)` or some other failure state to ensure that.

compare_platforms/compare_platforms_read_only.c
```diff
@@ -0,0 +1,209 @@
...
+    
+    print_details(&review);
+    
+    bool b_improved = false;
+    
+    char *newpublicreview = malloc(biggersz);
+    strcpy(newpublicreview, "5 ***** strong accept. This author deserves two Nobels and an ice cream.\n");
+
+#ifdef __CHERI_PURE_CAPABILITY__    
+    if(can_write(&review) == false)
```
0152la, line 177: What does this guard check against? Or more specifically, does the value of the boolean ever change? If so, how does that relate to the steps in the preamble?

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in d032455c2ca8a57911b597af27fb082fab8f2fd1

compare_platforms/compare_platforms_read_only.c
tel3path: can_write() checks if the review's permissions have a 'w' in them. 
The value of the boolean changes if rewrite_permissions_by_overflow() succeeds in adding one or more 'w's to the permissions.
Then the text of the review can be changed.

tel3path:
> 

Sorry, you are correct, I meant to reply to 

> > Asserts refactored to make sense according to the relative positions of the pointers: there should be no overlap, and the overflow-by-1 should come from the pointer on the immediate left of permissions. Fixed in [fa66ce0](https://github.com/capablevms/cheri-examples/commit/fa66ce07d959169de7f32c137822df458d99bc80)
> 
> I don't quite get this... is this relate to any of the conversation we had in this PR? Also curious if this affects our previous conversation [1] about overflow.
> 
> [1] [#78 (comment)](https://github.com/capablevms/cheri-examples/pull/78#discussion_r1083841299)

Sorry, you are correct, I meant to reply to [#78 (comment)](https://github.com/capablevms/cheri-examples/pull/78#discussion_r1083841299).

ltratt:
Something weird is happening to your comments, @tel3path, and they're disassociated from the comments attached to code, which makes following this very difficult. Can you work out why that might be? Thanks!

compare_platforms/compare_platforms_read_only.c
tel3path: Test

tel3path: Test

tel3path: In [#78 (comment)](https://github.com/capablevms/cheri-examples/pull/78#issuecomment-1404988563) I wrote:

"Asserts refactored to make sense according to the relative positions of the pointers: there should be no overlap, and the overflow-by-1 should come from the pointer on the immediate left of permissions. Fixed in fa66ce07d959169de7f32c137822df458d99bc80"

===

The review struct now has three members of two different sizes: smallsz and biggersz. The bigger-sized pointer is publicreview.

The order of the pointer addresses is different on my local machine (Apple clang), on bencher14 natively (Debian), and on morello-purecap. 

On Apple clang, the biggest pointer has the lowest address. 
On morello-purecap, the biggest pointer has the highest address.
On Debian, the biggest pointer is in the middle.

Previously, the asserts were not taking the different pointer sizes into account. The asserts check that the pointers are (expected to be) non-overlapping according to their allocated sizes. To do this, we have to check the order of the addresses.

The two smaller pointers are realname and permissions. 

In all cases, the permissions are initialized to r. Then the overflow tries to add one or more 'w's.

—

On Apple clang, the order is: the biggest pointer (publicreview) has the lowest address, followed by realname, followed by permissions. 

Because permissions comes after realname, the overflow-by-1 has to come from realname. This is the output:

Overflowing realname into permissions by 1:
Permissions: rw

—

On Debian, realname has the lowest address, followed by the biggest pointer, followed by permissions. 

Because permissions comes after the biggest pointer, the overflow-by-1 therefore has to come from the biggest pointer, publicreview:

Overflowing publicreview into permissions by 1:
Permissions: r

Here, the overflow-by-1 isn't enough to change the permissions.

—

On morello-purecap, realname has the lowest address, followed by permissions, followed by the biggest pointer. 

Here, permissions comes after realname, so the overflow-by-1 comes from realname, just as it did on Apple; but the outcome is a SIGPROT:

Overflowing realname into permissions by 1:
In-address space security exception (core dumped)

tel3path:
> Something weird is happening to your comments, @tel3path, and they're disassociated from the comments attached to code, which makes following this very difficult. Can you work out why that might be? Thanks!

I applied an update to my browser and cleared the cache. I hope that's all it is.

compare_platforms/compare_platforms_read_only.c
0152la: Right, but I see `rewrite_permissions_by_overflow` is not being called until this point.

compare_platforms/compare_platforms_read_only.c
0152la: Alright, that's a lot to unpack. If I understand correctly, this part of the code aims to overwrite a particular read flag to make it writable, by using *some* initial pointer. The issue you observed is that the flag's location is dependent on the struct's in-memory arrangement, which might be different based on local configuration.

So what fa66ce07d959169de7f32c137822df458d99bc80 is attempting to solve is identifying the offset of the flag we want to override from the *some* initial pointer we have?

If all of this is correct, I'd argue this goes against the original intent, which the preamble specifies as "try to exceed the bounds of an object and add write permissions (which will also SIGPROT).". For **this** PR, presuming that the functionality was correct before fa66ce07d959169de7f32c137822df458d99bc80, I would suggest we retract that commit, and merge as is (after CI checks and so on). We can then discuss the refinement you propose, if it's worth adding in a subsequent PR.

compare_platforms/compare_platforms_read_only.c
tel3path: Sorry, I didn't properly answer the question about how it relates to the steps in the preamble. On line 164, permissions are set to "r". Then we reach this CHERI block where on line 180, it checks the review struct's permissions before setting them to read-only.

With this program flow, the permissions are only ever going to be "r", so arguably this check is not necessary for the purpose of an example. Shall I remove it?

0152la: Yes, if the check doesn't do anything, best to remove it.

compare_platforms/compare_platforms_read_only.c
tel3path: Okay, that makes sense. I was concerned about leaving too much to chance, but in practice it was working, so I will retract that commit.

0152la:
One other thing. Make sure to add this test to the CI script. Presumably they should only be ran in purecap, so add them in the respective group [1]. I also don't think that script can handle arguments passed to the tests themselves, so let's just add the succeeding variant for now, without the flag.

[1] https://github.com/capablevms/cheri-examples/blob/master/tests/run_tests.sh#L60

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in 879f3f56213110b07385348d0802b4a99a53539d

compare_platforms/compare_platforms_read_only.c
tel3path: Fixed in e55ffa2458505e0a1149a60fe0a421e4da07bca7

tel3path:
> One other thing. Make sure to add this test to the CI script. Presumably they should only be ran in purecap, so add them in the respective group [1]. I also don't think that script can handle arguments passed to the tests themselves, so let's just add the succeeding variant for now, without the flag.
> 
> [1] https://github.com/capablevms/cheri-examples/blob/master/tests/run_tests.sh#L60



> One other thing. Make sure to add this test to the CI script. Presumably they should only be ran in purecap, so add them in the respective group [1]. I also don't think that script can handle arguments passed to the tests themselves, so let's just add the succeeding variant for now, without the flag.
> 
> [1] https://github.com/capablevms/cheri-examples/blob/master/tests/run_tests.sh#L60

Fixed in df5a0fe6c08bf20e0e2d5456fecd17e2854b378e

0152la:
I think that's more or less it from me. I don't know if @ltratt has anything else to add. While we wait for confirmation, could do some `bors try` to stamp out any potential CI issues.

ltratt:
bors try

ltratt:
I have a dumb question: is `compare_platforms` the right name for this example?

tel3path:
> ## try
> Build failed:
> 
> * [buildbot/ci-builder](http://bencher14.soft-dev.org:8010/#builders/2/builds/86)

Fixed in fbd56bca48818b916a983bd3fcb7188c3a972359

0152la:
What's the status of this PR?

tel3path:
> 

Waiting for bors try: https://github.com/capablevms/cheri-examples/pull/78#issuecomment-1408400326

I understand only reviewers can do that, or am I mistaken?

0152la:
Only reviewers can merge (`r+`), trying is free. Please keep trying until the CI succeeds.

ltratt:
Anyone can run bors try (but *only* reviewers can do `r+`).

tel3path:
bors try

tel3path:
bors try

tel3path:
bors try

tel3path:
bors try

tel3path:
bors try

ltratt:
Let's first `git revert` https://github.com/capablevms/cheri-examples/pull/78/commits/21bd0dfefd2a2b13a9842a5747331d926892def3. Then, unfortunately (but it's no-one's fault!) there's a merge conflict in `run_test.sh`. That'll require a manual merge with `git merge`. Push that, do a "bors try" and ... cross your fingers :)

tel3path:
bors try

ltratt:
@0152la I'm happy with this if you are.

0152la:
I'm happy as well; please squash.

ltratt:
bors try

tel3path:
> Uh, did this somehow get merged without bors?

My fault. I tried various ways of squashing, none of which were correct. Then I tried the squash button in the browser, not realizing it was the wrong thing to do.

0152la:
Didn't realize there was a squash button in the browser; I apologize for not being more elaborate. It seems it more or less *looks* like what we want, but the commit message is particularly ugly. I can help try untangle this tomorrow if you'll still be needing help.

tel3path:
> Didn't realize there was a squash button in the browser; I apologize for not being more elaborate. It seems it more or less _looks_ like what we want, but the commit message is particularly ugly. I can help try untangle this tomorrow if you'll still be needing help.

Thank you. At this point I think the rational choice is to sleep on it.

ltratt:
I *think* all that's needed is probably:

```
git fetch --all
git rebase -i upstream/master
```

That should rebase this and get rid of the remnants of the merge conflict. If the `rebase` fails, please let me know. If it succeeds, please force push.

tel3path:
> I _think_ all that's needed is probably:
> 
> ```
> git fetch --all
> git rebase -i upstream/master
> ```
> 
> That should rebase this and get rid of the remnants of the merge conflict. If the `rebase` fails, please let me know. If it succeeds, please force push.

The `rebase` failed again I'm afraid:

`git rebase -i upstream/master'

`Auto-merging tests/run_tests.sh`

`CONFLICT (content): Merge conflict in tests/run_tests.sh
error: could not apply 0d366da... changed cap_to_file and leak-capability from passing to failing`

`hint: Resolve all conflicts manually, mark them as resolved with`

`hint: "git add/rm <conflicted_files>", then run "git rebase --continue".`

`hint: You can instead skip this commit: run "git rebase --skip".`

`hint: To abort and get back to the state before "git rebase", run "git rebase --abort".`

`Could not apply 0d366da... changed cap_to_file and leak-capability from passing to failing`

tel3path:
bors try

ltratt:
bors try-

ltratt:
bors ping

ltratt:
@tel3path I think you might have to try reopening this PR and then we can try bors. [If that doesn't work we'll have to open a new PR.]

tel3path:
> @tel3path I think you might have to try reopening this PR and then we can try bors. [If that doesn't work we'll have to open a new PR.]

Looks like I can't reopen it. Will have to open a new one.

ltratt:
OK.

tel3path:
> 

When I tried to push, it was rejected:

> error: failed to push some refs to 'https://github.com/tel3path/cheri-examples'
> hint: Updates were rejected because the tip of your current branch is behind
> hint: its remote counterpart. Integrate the remote changes (e.g.
> hint: 'git pull ...') before pushing again.
> hint: See the 'Note about fast-forwards' in 'git push --help' for details.

When I tried to pull, I got:

> hint:   git config pull.rebase false  # merge
> hint:   git config pull.rebase true   # rebase
> hint:   git config pull.ff only       # fast-forward only

Which of these pull commands do you think I should do?

ltratt:
I'm not quite sure what you're trying, but maybe an easy alternative is to do something like:

```
git checkout -b compare_platforms
git push --set-upstream origin compare_platforms
```

and then raise a PR as normal.

PR #83, opened 2023-02-23
Add an example to compare a stack smash in CHERI purecap vs. morello-hybrid vs. non-CHERI environments.

tel3path:
The purpose of the example is to compare how a stack-smashing attempt is handled in a CHERI purecap vs hybrid environment, vs a non-CHERI environment.
 
The example first displays a substring of the Latin alphabet starting from a random index, 
then tries to smash the stack using an index that is known to be out-of-bounds, 
and display the out-of-bounds substring to the user.

The example can be run as a single instance, but is meant to be run in several instances launched simultaneously, to compare how the respective environments handle the pointer addresses.
 
 In a CHERI purecap environment:
 - A single instance should crash, and multiple instances should hang;
 - The user should not see the out-of-bounds substring;
 - The address of the pointer should be the same for all instances but change when
 the substring's value changes.
 
 In a morello-hybrid environment:
 - The program should run without crashing;
 - The user should see the out-of-bounds substring;
 - The address of the pointer should be different for all instances and should not change when the substring's value changes. 
 
In a non-CHERI environment, the program's behaviour is implementation dependent.


tel3path:
bors try

tel3path:
bors try

compare_platforms/compare_platforms_stacksmash.c
```diff
@@ -0,0 +1,109 @@
+/*
+ * This example is meant to be run in multiple instances launched simultaneously.
+ * It can also be run as a single instance.
+ *
+ * Usage:
+ * prompt> ./compare_platforms_stacksmash & ./compare_platforms_stacksmash &
```
ltratt, line 6: I must admit that I don't understand why this needs two instances running side by side to be a problem -- this process doesn't seem to interact with the outside world in such a way that would cause a problem? Perhaps I'm missing something!

compare_platforms/compare_platforms_stacksmash.c
```diff
@@ -0,0 +1,109 @@
...
+    const size_t twice_alphabet_size = 52;
+    size_t evilindex = twice_alphabet_size - goodindex;
+    // change the message to something the user shouldn't see
+    str = &alphabet[evilindex];
+
+#ifdef __CHERI_PURE_CAPABILITY__
+    printf("\nPID = %d\n", getpid());
+    pp_cap(str);
+#else
+    printf("\nPID = %d\n", getpid());
```
0152la, line 97: The first `printf` seems identical; can this be lifted out of the `ifdef`?

compare_platforms/compare_platforms_stacksmash.c
```diff
@@ -0,0 +1,109 @@
...
+    printf("Address of str = %p\n", &str);
+#endif
+
+    // keep showing the ill-gotten string until the user kills the process
```
0152la, line 101: Is this an intentional non-terminating loop? I would advise against having this. Is there some reason we don't terminate?

0152la (commented):
Can you please indicate more precisely where the stack comes into play? Is it because we expect `alphabet` and `evil message` to reside on the stack consecutively?

compare_platforms/compare_platforms_stacksmash.c
tel3path: My bad; the loop terminates after 10 repetitions, but I forgot to update the comment to reflect that. Fixed in fb282b0443b2d7feb478db5ade8b3f9597ce0277.

compare_platforms/compare_platforms_stacksmash.c
ltratt: Yes, I think the multiple processes thing is a red herring, so it's best removed.

compare_platforms/compare_platforms_stacksmash.c
```diff
@@ -0,0 +1,109 @@
...
+    struct timeval t;
+    intptr_t rc = gettimeofday(&t, NULL);
+    assert(rc == 0);
+    return (intmax_t) t.tv_sec + (intmax_t) t.tv_usec / 1e6;
+}
+
+void spin(intptr_t howlong)
+{
+    intmax_t t = get_time();
+    while ((get_time() - t) < (intmax_t) howlong)
```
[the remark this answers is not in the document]
tel3path: Fixed in 1cb94160c5c7f6b2703f5ab468969701039880d2

compare_platforms/compare_platforms_stacksmash.c
tel3path: Fixed in 8a6214b7e2c153dfe3c8fa47db66cef579c07964

tel3path:
> Can you please indicate more precisely where the stack comes into play? Is it because we expect `alphabet` and `evil message` to reside on the stack consecutively?

In a nutshell, yes. It is the printf() that results in the SIGPROT caused by a capability bounds fault. Have added an explanatory comment to this effect.

Fixed in 42de83e046b476a63832eeb0cc60a61e511c6b70.

compare_platforms/compare_platforms_stacksmash.c
tel3path: Fixed in 632acd64b8eee4b438b2b4bc36faa5507f61eb54

tel3path:
bors try

0152la:
As discussed offline, the changes in this PR in its current form don't seem to be too substantially different from existing examples, so we'll close this for now.

PR #88, opened 2023-11-06
Example allocator making compartments on the heap

probablytom:
An unsophisticated hybrid allocator which create compartments on the heap, and allows `malloc`s against those compartments.

This is a modification of the existing bump allocator which works in hybrid mode and demonstrates these little compartments which — in theory — could sit "to the side" of the typical DDC-enforced compartments we use to police memory at the moment.


ltratt:
Can we hook this into the build system so that buildbot checks it? @0152la might know the best way of doing that.

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+    if (new_allocated > compartment.max_allocated) {
+        return 0;
+    }
+
+    compartment.bytes_allocated = new_allocated;
+
+    // We allocated some space!
+    // Create a capability pointing to it and return it.
+    // The capability is bounded by the data capability for this compartment.
+    allocated = cheri_cap_build((void *__capability)addr, (__uintcap_t)compartment.datacap);
```
ltratt, line 86: I must admit that I don't think I've seen `cheri_cap_build` before -- I think it's new? Though Yury has used it in his blog post, I see!

ltratt:
You'll probably want to run LLVM's C formatter on this *but* wonderfully that formatter isn't entirely stable. By convention (since it's what we use in buildbot) we use the Morello C formatter https://github.com/capablevms/cheri-examples/blob/master/.buildbot.sh#L12

example_allocators/compartment_alloc/README.md
```diff
@@ -0,0 +1,17 @@
+# Compartmentalising Allocator
+
+A modification of the "bump allocator" in this repo.
```
ltratt, line 3: Minor nitpicking. Normally we format text files to 80 chars.

0152la:
> Can we hook this into the build system so that buildbot checks it? @0152la might know the best way of doing that.

It would have to be added somewhere appropriately in the test file [1]. Presuming in the `morello-hybrid` block skimming over the comments.

[1] https://github.com/capablevms/cheri-examples/blob/master/tests/run_tests.sh#L60-L93

example_allocators/compartment_alloc/README.md
```diff
@@ -0,0 +1,17 @@
...
+This allocator creates compartments on the heap, and returns capabilities
```
0152la, line 4: More of a philosophical question than asking for a specific change, but what does `compartments` mean in this instance?

0152la:
Also you'll need to run `clang-format` [1] to have it be accepted by our CI.

[1] https://github.com/capablevms/cheri-examples/blob/master/.buildbot.sh#L9-L13

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+
+/*
+ * Initialise a compartment with `size_in_bytes` space, where all future
+ * allocations will be constrained by `dc` as if it was the DDC when allocation
+ * occurred.
+ */
+void
+*__capability init_compartment(int size_in_bytes, void *__capability dc)
+{
+    // We impose a limit on our # of compartments
```
0152la, line 20: Why do we impose a limit?

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+heap_compartment compartments[maxCompartments];
+int numCompartments = 0;
+
+/*
+ * Initialise a compartment with `size_in_bytes` space, where all future
+ * allocations will be constrained by `dc` as if it was the DDC when allocation
+ * occurred.
+ */
+void
+*__capability init_compartment(int size_in_bytes, void *__capability dc)
```
0152la, line 18: I would make `size_in_bytes` unsigned.

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
+#include <cheriintrin.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include "compartment_alloc.h"
+#include "utils.h"
+
+heap_compartment compartments[maxCompartments];
```
0152la, line 9: Just double checking this does get initialized automatically because it's a global, and we don't have to explicitly `malloc` it.

example_allocators/compartment_alloc/main.c
```diff
@@ -0,0 +1,29 @@
...
+  // Print an explainer of each allocation (and the capability itself)
+  printf("\n\n\tThis capability is heap-allocated within our first compartment, which is write-only.\n\tWe allocated 64 bytes.\n");
+  pp_cap(c1_allocated_memory);
+
+  printf("\n\n\tThis capability is heap-allocated within our second compartment, which is read-only.\n\tWe allocated 256 bytes.\n");
+  pp_cap(c2_allocated_memory);
+
+  printf("\nCompleted successfully.\n");
+
+  return 0;
```
0152la, line 28: I believe there is leaked memory at this point. Perhaps a `clean_compartment` function or so to `free` all the `mmap`s in `init_compartment`.

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+{
+    size_t rounded_length;
+    size_t new_allocated;
+    heap_compartment compartment = compartments[0];
+    void *__capability allocated;
+    void *addr;
+    int i;
+
+    // Search for a compartment with the given identifier
+    for (i = 0; i < maxCompartments && compartments[i].identifier != compartment_id; i++){}
```
0152la, line 60: Not quite sure what's happening here. Is there meant to be no semicolon after the empty body of this for? Can't we omit the `{}` completely and just leave the for itself?

Also, did you check that edge case if the sought-for compartment is at the end of `compartments`?

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+
+    // Try to "bump-allocate" some space in the compartment's buffer
+    addr = compartment.buffer + compartment.bytes_allocated;
+
+    addr = __builtin_align_up(addr, ~cheri_representable_alignment_mask(len) + 1);
+    rounded_length = cheri_representable_length(len);
+
+    new_allocated = (addr + rounded_length) - compartment.buffer;
+    if (new_allocated > compartment.max_allocated) {
+        return 0;
```
0152la, line 78: I think there should be some more effect (perhaps even an outright error) if we allocate more than we have left.

example_allocators/compartment_alloc/compartment_alloc.c
0152la: Is there some documentation somewhere for what this function is meant to do? It might be best to just cast to a capability to avoid any potential side-effects if we're not sure what this is meant to accomplish.

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+        exit(-1);
+    }
+
+    // Create the new compartment
+    compartments[numCompartments] = (heap_compartment){
+        buf,                     // Space on the heap
+        0,                       // bytes allocated (currently none!)
+        size_in_bytes,           // Our maximum size, to check when allocating
+        dc,                      // The data capability to allocate against (like a DDC)
+        sealed_reference(buf)    // An identifier which allows a holder to allocate in this component
```
0152la, line 39: Where is `sealed_reference` coming from?

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+
+    // Return the sealed capability which acts as a token of authentication for a component.
+    return compartments[numCompartments++].identifier;
+}
+
+/*
+ * Allocate `len` bytes in the compartment identified by `compartment_id`.
+ * if `compartment_id` doesn't exist, returns NULL
+ */
+void *__capability malloc_compartment(int len, void *__capability compartment_id)
```
0152la, line 50: I would also make `len` unsigned.

example_allocators/compartment_alloc/compartment_alloc.c
```diff
@@ -0,0 +1,91 @@
...
+    compartments[numCompartments] = (heap_compartment){
+        buf,                     // Space on the heap
+        0,                       // bytes allocated (currently none!)
+        size_in_bytes,           // Our maximum size, to check when allocating
+        dc,                      // The data capability to allocate against (like a DDC)
+        sealed_reference(buf)    // An identifier which allows a holder to allocate in this component
+    };
+
+    // Return the sealed capability which acts as a token of authentication for a component.
+    return compartments[numCompartments++].identifier;
```
0152la, line 43: Is this meant to also increment `numCompartments`? Personally I would put it on a separate line, as this effect might be overlooked, but I don't feel too strongly about it.

0152la (changes requested):

example_allocators/compartment_alloc/README.md
probablytom: Great Q…

Here, the "compartment" is some heap-allocated space that only capabilities are handed out to. There are obviously ways to abuse "just some space on the heap" if the DDC isn't also restricted. The easiest way to explain this — I think — is to draw it out, so please excuse the long response…

I've begun to draw out compartments when I think about them like this. Assuming only the DDC is being used for compartmentalisation (so we're constraining, say, dereferencing of addresses for data but not jump points etc) and it's being used to bound which addresses can be legally dereferenced, we could draw memory out as a "number line" of the address space and mark points where a compartment's bounds begin and end. So we could diagram compartments like so:

```
|--------------------|
0  g   f      F   G  ♾️
```

where a compartment's lower bound is a lowercase letter and its upper bound the corresponding upper case letter (here ♾️ is the maximum address, and the `G` compartment's lower bound is marked by `g`, and the upper case marked by `G`). I'm obviously just defining compartment bounds relative to each other, and not in any kind of detail.

The compartments this example constructs are intended to sit _outwith_ all other compartments, so that any access performed against them _must_ be done through the capabilities returned by `malloc_compartment`. For this to work, obviously the DDC needs to be restricted when a program first runs, so that no future compartment's bounds will overlap with the ones we construct on the heap. Then, the ordinary compartments we create (derived from the now-restricted initial DDC) can't overlap with these separated heap-y compartments. I've taken to thinking of them as "side compartments" to differentiate them from ordinary compartments. To illustrate what I mean, we can create a special space adjacent to our ordinary compartments and allocations like this:

```
|---------------------|
0              ><     ♾️ 
```

Where `>` marks the upper bound of a DDC we install right when a program launches, and `<` marks the lower bound of possible "side" compartments. To extend the first diagram, we might make a side compartment that looks like…:

```
|------------------------------|
0  g   f    F   G  ><  p   P   ♾️
```

…and the "side" compartment `P` is what `compartment` means in this instance. Those compartments are similar to the buffers created in the `bump-allocator` example in structure, but this is long enough so I'll leave their definition in the code to speak to that. There's a little work to do, I suppose, depending on how thorough the example aims to be, but this is just a toy to demonstrate a kind of cheri-aware hybrid allocator.

Apologies for the wall of text, hope this is clear enough to explain what I mean :)

probablytom: (I've got a similar but slightly more detailled write-up of compartment structure and these "side" compartments I wondered about adding to [the notes on compartments that already exist](https://github.com/capablevms/cheri_compartments), in case this concept ends up being useful and the definition wants to live somewhere…)

example_allocators/compartment_alloc/main.c
probablytom: Sure! I can add that. It'll need to do pointer revocation on the capabilities handed out to the memory being cleared, right? I haven't seen examples of that yet, if you have a code snippet handy I'd be curious to see (no worries if not, I'll google and implement when I get a sec).

(FWIW, this is following the lead of the bump allocator, which just has two functions analogous to the two this example contains — maybe I'll extend that too if I get a slice of time for it.)

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: It's intended to create a capability with the metadata of `compartment.datacap` (bounds, permissions etc) and the address of `addr`. For some reason it requires the address as a capability and doesn't seem to set it properly in this example…I've got a snippet where it works as expected and never solved the mystery here. It needs tidying up. I didn't realise it was uncommon, I'll change this to be more idiomatic!

The result I wanted to achieve with these three lines was to have a capability with the address `addr`, bounds set to exactly the bounds of the allocated # bytes, and all other data derived from `compartment.datacap` (I set r/w permissions in `main.c`, for example). Is there an idiomatic way to construct that capability without using `cheri_cap_build()`?

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Ah rats! It's coming from `utils.h`, which I completely missed when adding files to the PR. Sorry…!

Fixed in [0ab49e8](https://github.com/capablevms/cheri-examples/pull/88/commits/0ab49e82383482093c8ba8857525f4777daba621).

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: 🙏 fixed in [912aa72](https://github.com/capablevms/cheri-examples/pull/88/commits/912aa728214fd1cea9fb6f8cb5ec298278c5c016)

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: 🙏 fixed in [912aa72](https://github.com/capablevms/cheri-examples/pull/88/commits/912aa728214fd1cea9fb6f8cb5ec298278c5c016)

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Yes, I can see that — thanks. Fixed in [c362c4a](https://github.com/capablevms/cheri-examples/pull/88/commits/c362c4ac0f35c2ef8ded36fa8371243ea7c0328f).

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Just because we've allocated a limited number in the `compartments` array, no other reason. Maybe I should remove the comment…?

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Yep: https://stackoverflow.com/a/16015741

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: An empty body and a semicolon immediately following the `for` brackets will do the same thing — I'm not sure what the idiomatic C would be here, so happy to follow whatever our convention is for something like this. Would you prefer to see `for (i = 0; i < maxCompartments && compartments[i].identifier != compartment_id; i++);`, or would you rewrite it another way?

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Yes! And I understand — made the change, fixed in [b84412b](https://github.com/capablevms/cheri-examples/pull/88/commits/b84412b1a51c116b237d5ebe1b00e636bbd2cf05).

example_allocators/compartment_alloc/README.md
probablytom: Fixed in [9c09bab](https://github.com/capablevms/cheri-examples/pull/88/commits/9c09babd299c9b6996d49a2ecfc63647a12e528c)!

example_allocators/compartment_alloc/compartment_alloc.c
ltratt: Can you use something like `cheri_address_set(compartment.datacap, addr)`? Warning: I might be wrong.

example_allocators/compartment_alloc/compartment_alloc.c
ltratt: Yeah, we might as well remove the comment. For an example we don't need to justify *why* we're being restrictive: it's an example :)

example_allocators/compartment_alloc/main.c
ltratt: For an example we can just put a comment along the lines of "// In the interests of simplicity we allow the XYZ to leak." if it's not worth writing a clear-up routine. [I don't feel strongly either way.]

example_allocators/compartment_alloc/compartment_alloc.c
0152la: I would just cast to `__capability` (and then you'll need `__from_cap` I think - you should get a warning) instead of `cheri_cap_build`, which I guess we don't know where it comes from?

example_allocators/compartment_alloc/compartment_alloc.c
0152la: Yeah, probably best to remove it here for the example, or at least give some justification if you want to keep it.

example_allocators/compartment_alloc/main.c
0152la: >  It'll need to do pointer revocation on the capabilities handed out to the memory being cleared, right?

I'm not sure where pointer revocation comes from. I just meant to clean up the `mmap`s in `init_compartment` as it's considered a memory leak to not do so, to my knowledge (could verify with `asan` or so).

example_allocators/compartment_alloc/compartment_alloc.c
0152la: It just looks confusing to me. If you think it's fine, leave as is. I just prefer being more verbose.

ltratt: There's no great answer here -- personally, I think the `{}` is the least worst option (it's what I do in this situation).

example_allocators/compartment_alloc/README.md
probablytom: I've now added the above notes to the README in [da50830](https://github.com/capablevms/cheri-examples/pull/88/commits/da5083095393c676bf679fb4e5368fef8e4ae83c).

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Removed in [c37081b](https://github.com/capablevms/cheri-examples/pull/88/commits/c37081bbcec4a81ff86cb083260dbddca70dc80d), thanks both!

example_allocators/compartment_alloc/main.c
probablytom: Compartments are now freed, fixed in [bd783eb](https://github.com/capablevms/cheri-examples/pull/88/commits/bd783eb6394493b22697536ff14049ca9b5a2e2a) — how does that look?

example_allocators/compartment_alloc/compartment_alloc.c
probablytom: Moved to `cheri_address_set` in [3c39dc7](https://github.com/capablevms/cheri-examples/pull/88/commits/3c39dc77ae584f060b5cc633c019f3979695adf4)

ltratt:
Please squash.

example_allocators/compartment_alloc/compartment_alloc.c
0152la: We did some investigation around `cheri_cap_build` and, in the initial usage, it was a no-op. We will likely have an example to submit with `cheri_cap_build` usage, which @probablytom will take care of, just to show how it can be used. We also found a comment from @jacobbramley that explains how this function should be used (as documentation on it seems severely lacking).

[1] https://github.com/capablevms/cheri-examples/pull/39#discussion_r714883024

0152la (approved):

probablytom:
Squashed (sorry for the noise). From the other PRs it looks like `bors` is responsible for actually merging — does it matter who invokes it?

0152la:
We just deprecated bors, and reviewers are meant to decide when to merge (and for the future, when to allow force pushes to be made).

0152la:
@probablytom Could you please amend the commit message to be one consistent overview, rather than the squashed commit concatenation? Just discuss the overall changes, and omit small, uninteresting changes due to the review process.

probablytom:
Done!

PR #92, opened 2023-11-14
Example comparing`cheri_cap_build` & `cheri_address_set`

probablytom:
`cheri_cap_build` and `cheri_address_set` are similar, but not quite the same. We were previously unsure of what the difference between the two was; this example uses both and explains the differences between the two.

It's pretty explicit about what's going on and prints detailled messages describing the capabilities it constructs & prints. I figured it'd be useful to have some detail in there so we didn't have to re-parse the code every time we wanted to remember what the difference between the functions is.

Put together mostly by @0152la; I just cut down the program we were using as a scratchpad when figuring this out and tidied it into an example.

### DIFF
--- a/cap_build.c
+++ b/cap_build.c
@@ -1,0 +1,159 @@
+// We've found little documentation of `cheri_cap_build`, and on the surface,
+// it seems similar to `cheri_address_set`. These functions do similar things,
+// but they're different. This example tries to demonstrate what those
+// differences are. Briefly summarised…:
+//
+// - `cheri_address_set()` takes a cap and a ptr, and returns a new cap with
+//    the address set to the ptr passed in, derived from the cap passed in.
+//    It's basically taking an address and popping it into the appropriate
+//    place in an existing capability to create a new one.
+// - `cheri_cap_build()` takes a cap and a uintcap_t (a uintptr_t equivalent
+//    for capabilities) and returns a new cap with the bits of the uintcap_t,
+//    which is valid if and only if it would be no more permissive than the cap
+//    passed as the first argument. This is taking all of the data in the
+//   `uintcap_t` and creating a cap from it, rather than deriving any of it
+//    from the first argument --- which differentiates `cheri_address_set` and
+//   `cheri_cap_build`.
+
+#include "cheriintrin.h"
+#include "include/common.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+int main()
+{
+    // This example relies on morello-specific symbols, so we won't run
+    // it on anything else.
+#if !(defined(__aarch64__) && __ARM_ARCH == 8)
+    printf("Skipping; this example only runs on morello.");
+#else
+    // We create two capabilities: x, and y, which is derived from x. We note
+    // that, though we use `cheri_address_set()`, it doesn't actually set (i.e.
+    // _change_) the address of its arguments, but construct a new capability
+    // with the address we pass.
+    //
+    // cheri_address_set(void* __capability, uintptr_t) -> void *__capability
+    void *__capability x = cheri_ddc_get();
+    uintptr_t y_ptr = (uintptr_t) malloc(20);
+    void *__capability y = cheri_address_set(x, y_ptr);
+
+    // The second capability printed below is created from the first using
+    // `cheri_address_set()`. Notice that `cheri_address_set()` created a new
+    // capability rather than setting the address on the first.
+    pp_cap(x);
+    pp_cap(y);
+    printf("\n\n");
+
+    // We create an (invalid!) capability by writing it directly.
+    void *my_cap_mem = malloc(16);
+    long cap;
+    cap = 0xdc5f400000030005;
+    memcpy(my_cap_mem + 8, &cap, 8);
+    cap = 0x00000000deadbeef;
+    memcpy(my_cap_mem + 0, &cap, 8);
+    void *__capability my_cap = (void *__capability) *(void *__capability *) my_cap_mem;
+
+    // We build a new capability using `cheri_cap_build()`. Note that this
+    // capability is _valid_, because it's derived from `x`, which is valid.
+    // The address comes from `my_cap`, which isn't valid, but we can still
+    // create a valid capability beecause `cheri_cap_build` just reads it's
+    // address bits and constructs a capability.
+
+    // To date we haven't found the specific type signature for this, but at
+    // https://www.morello-project.org/resources/morello-linux-morelloie/
+    // we can see:
+    // void *__builtin_cheri_cap_build(const void *key, unsigned __intcap bits);
+    // After experimenting, we've found it wants to be treated as…:
+    // cheri_cap_build(void *__capability, uintcap_t) -> void *__capability
+    void *__capability valid_cap = cheri_cap_build(x, (uintcap_t) my_cap);
+
+    // The first capability printed below is created by writing its structure
+    // into memory; it's not valid, it's just a bunch of bits. The second is an
+    // equivalent capability, which is constructed using `cheri_cap_build().
+    // Notice that the second one is valid; that's because it's derived using a
+    // valid capability (the variable `x`, which contains a copy of the DDC).
+    pp_cap(my_cap);
+    pp_cap(valid_cap);
+    printf("\n\n");
+
+    // It might appear that `cheri_cap_build` could be used to create
+    // capabilities with elevated permissions, by writing specific permission
+    // bits that we want to elevate before recreating a capability. That's not
+    // possible. To demonstrate, the rest of this example uses
+    // `cheri_cap_build` to rebuild capabilities from raw bits stored in the
+    // heap.
+
+    // Write a copy of the DDC to heap, so we can play with it later
+    char *ddc_copy_space = malloc(16);
+    void *__capability ddc_copy_val = cheri_ddc_get();
+    void **ddc_copy_raw_ptr = (void **) &ddc_copy_val;
+    memcpy(ddc_copy_space, ddc_copy_raw_ptr, 16);
+
+    // Switch DDC, removing the mutable load permission, so we can attempt to
+    // restore permissions we _shouldn't_ have later on.
+    void *__capability newDDC = cheri_perms_and(cheri_ddc_get(), ~ARM_CAP_PERMISSION_MUTABLE_LOAD);
+    asm("MSR DDC, %[cap]\n\t" : : [cap] "C"(newDDC) : "memory");
+
+    // Create a capability and copy it into heap; use this to construct two new
+    // capabilities, one unmodified and one with restored permissions, both
+    // using cheri_cap_build We remove some permissions from the DDC and try to
+    // construct a capability that has them again.
+    int val = 5;
+    char *modspace = malloc(16);
+    void *__capability val_cap = (void *__capability) &val;
+    void **cap_bytes_raw_ptr = (void **) &val_cap; // a pointer to the cap's bytes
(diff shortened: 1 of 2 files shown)


cap_build.c
```diff
@@ -0,0 +1,68 @@
...
+    // cheri_address_set(void* __capability, uintptr_t) -> void *__capability
+    void *__capability x = cheri_ddc_get();
+    uintptr_t y_ptr = (uintptr_t) malloc(20);
+    void *__capability y = cheri_address_set(x, y_ptr);
+
+    printf("\n\tThe second capability printed below is created from the first using");
+    printf(
+        "\n\t`cheri_address_set()`. Notice that `cheri_address_set()` created a new capability,");
+    printf("\n\trather than setting the address on the first.");
+    printf("\n\n");
```
0152la, line 22: I would change these prints into comments, so on executing, we just have the capability prints, and then can refer to side-by-side with the example code. It's more that this is how all examples (to my knowledge) are done.

cap_build.c
```diff
@@ -0,0 +1,68 @@
...
+    memcpy(my_cap_mem + 0, &cap, 8);
+    void *__capability my_cap = (void *__capability) *(void *__capability *) my_cap_mem;
+
+    // We build a new capability using `cheri_cap_build()`. Note that this capability is _valid_,
+    // because it's derived from `x`, which is valid. The address comes from `my_cap`, which isn't
+    // valid, but we can still create a valid capability beecause `cheri_cap_build` just reads it's
+    // address bits and constructs a capability.
+
+    // cheri_cap_build(void *__capability, uintcap_t) -> void *__capability
+    void *__capability valid_cap = cheri_cap_build(x, (uintcap_t) my_cap);
```
0152la, line 42: Can you explain why we need the `uintcap_t` cast here? I *think* it's because it's 128 bits, rather than a capability's 129, but not sure. You can also just say we need this since it works.

cap_build.c
```diff
@@ -0,0 +1,68 @@
...
+    printf("\n\tThese functions do similar things, but they're different.");
+    printf("\n\t`cheri_address_set()` takes a cap and a ptr, and returns a new cap with the");
+    printf("\n\taddress set to the ptr passed in, derived from the cap passed in.");
+    printf("\n\t`cheri_cap_build()` takes a cap and a uintcap_t, and returns a new cap with the");
+    printf("\n\taddress stored in the uintcap_t.");
+    printf("\n");
+    printf("\n\tWhile these are very similar, `cheri_cap_build()` is intended to be used when");
+    printf("\n\tyou're creating capabilities from \"raw data\". In general, if you're unsure");
+    printf("\n\twhich to use, it's likely `cheri_address_set()`.");
+    printf("\n\n");
```
0152la, line 63: I would either drop this whole discussion (or sprinkle it in comments appropriately), or have an example which emphasises the difference (which the rest of this example might do).

0152la (changes requested):

ltratt:
Since I'm the dummy here, I'll ask the obvious question: does `cheri_cap_build` allow you to create an arbitrary capability that overrides all other checks? I guess "no", but I'm not quite sure what the "but" is. My initial guess is that the capability that's created is valid but has limited permissions -- but maybe I'm wrong!

ltratt:
Thanks!

> I have a suspicion it can be used to reconstruct a capability that's been serialised to a file

This shouldn't be possible -- *but* there was once a bug in CheriBSD that allowed it because of file caching! It was fixed, though.

djlowther:
`cheri_cap_build` translates to the `BUILD` Morello instruction, which like Tom suspects turns a bit-pattern into a valid, unsealed capability, provided that that capability can be derived from the capability argument. I believe it's designed for situations like virtual memory paging -- the kernel can use it to quickly reconstitute capabilities in swapped pages rather than having to use a sequence of SCBNDS, SCVALUE, etc.

ltratt:
Thanks both -- I get it now!

cap_build.c
probablytom: Sorry for the delay. Fixed in [be6d139](https://github.com/capablevms/cheri-examples/pull/92/commits/be6d139256ad63fa4067d244d8082f8926efe9fd) — thanks!

cap_build.c
probablytom: We need this since it works 😛

More seriously though, the function only accepts a `uintcap_t`. You might remember when we were pairing on this that we got lots of cryptic errors from the compiler around the type of this argument. It wanted a `uintcap_t` — I imagine because it specifically doesn't care about the validity bit of its second argument (as it'll decide for itself whether the capability it returns should be valid), but it needs every other bit of the capability, because whatever it returns will be a capability with these 128 bits…just not necessarily a valid one.

cap_build.c
probablytom: I reworked it (with updates following @djlowther's explanation of `cheri_cap_build` from today — thanks!) and moved the relevant parts to the top of the example as an explanation of what the example is trying to demonstrate / accomplish and as a brief intro for anybody using it to learn. WDYT of the current incarnation? (Fixed in [a1b2ba1](https://github.com/capablevms/cheri-examples/pull/92/commits/a1b2ba1cc12effdf2fa5b0bce7d8ee3bf7b6ddd8))

probablytom:
Thanks @djlowther — that's a very helpful note! We haven't found much documentation around `cheri_cap_build` ourselves.

I haven't played with this yet and can't budget time to investigating it right now, but perhaps you know already: does that mean that a capability written to disk by one process could be loaded by a hybrid process with a very permissive DDC, and so re-created as a valid capability by that second process? Shouldn't that be forbidden, as @ltratt mentioned? Or am I misunderstanding something?

probablytom:
Fascinating! Thank you for the detailled response!

cap_build.c
0152la: I meant more in the code, as a comment. I recall we found some builtin function signature of this `cheri_cap_build` that took two identically typed arguments in the representation, but passing a capability as the second one gave an error (while obviously a capability as the first one is fine). 

If you, by chance, found some more details about the function signature of this, that would be great, if not, no worries. Can say we experimented and this seems to be the signature, just add that as a comment.

probablytom: Ah! Sorry, misunderstood. Closest to a definition I ever found was at the end of [this article from ARM](https://www.morello-project.org/resources/morello-linux-morelloie/). Added a note on that in [f0dbba6](https://github.com/capablevms/cheri-examples/pull/92/commits/f0dbba61699584ac1fa59191d2d7c276e2662669).

0152la: Right, Yury did mention when that article got uploaded. I wonder how related it is to his examples repo.

ltratt:
@0152la I think this is OK to squash?

0152la (approved):

probablytom:
Squashed!

0152la:
On my end, still says it's out of date with master.

probablytom:
Oh, and mine too. Apologies, thought I'd fixed — one sec

0152la:
I've removed it from the merge queue, since this now seems to have our whole repo history in it.

probablytom:
Yes, fortunately it looks like it was also still marked as out of sync with master so didn't auto-merge. Should be fixed now — looks correct on my end, at least — but will keep my mitts off the merge button…

0152la:
Forgot one thing; should we add this to CI? I would, but I'll defer to @ltratt if we want to merge as is, since it's more "this is how you use this" of an example.

ltratt:
Good point about adding it to CI. @probablytom Can you push a commit which does that please?

probablytom:
Hang on while I rebase again to resolve the conflict…

0152la:
LGTM

ltratt:
Please squash.

probablytom:
Sorry for the noise. Should be good now.

tests/run_tests.sh
```diff
@@ -50,7 +50,14 @@ function run {
             make -f Makefile.$BUILDBOT_PLATFORM run-$name
         fi
     done
+<<<<<<< HEAD
```
ltratt, line 53: Merge conflict left in. Please force push an update.

probablytom: 🤦

probablytom: Fixed in force-pushed [f509753](https://github.com/capablevms/cheri-examples/pull/92/commits/f509753a0b6fa89ee15be5499c747ccb018fb704).

ltratt:
Please squash.

probablytom:
…should be good to go… 👍

probablytom:
Re-formatted.

ltratt:
@probablytom Is this ready for rereview?

0152la:
I'm confused about commit https://github.com/capablevms/cheri-examples/pull/92/commits/ba68b10ed686829e326fb955857718d4cfa53486. Looking at the code [1], this new example should only run on `morello-hybrid`. If you see it running on `risc-v`, that might be a CI issue, and I need to investigate.

[1] https://github.com/capablevms/cheri-examples/pull/92/files#diff-bfa67bbd428f0fd40485bf6e666132a2adb30f6524631aff49d2bcd30f71c5ebL75

ltratt:
>

ltratt:
> I'm confused about commit https://github.com/capablevms/cheri-examples/commit/ba68b10ed686829e326fb955857718d4cfa53486. Looking at the code [1], this new example should only run on morello-hybrid. If you see it running on risc-v, that might be a CI issue, and I need to investigate.

I agree that this might be worth investigating, but I'd prefer to get this PR in now, and investigate separately: we can always reenable on RISC-V later if it turns out there's a CI issue.

@probablytom Just checking that this PR is finished from your perspective?

probablytom:
Yes, I was confused as to why it was running too. I think it's because the CI runs `make -f Makefile.riscv64-purecap all`, and the make rule appears to pick up _every_ example in the root of the repo and run it as if it were riscv-compatible. Other examples have similar macros to avoid running on the incorrect architectures [1], I imagine for exactly this reason — so I _think_ that commit [2] actually follows the precedent set by other examples so far. Afraid I haven't been able to investigate further, but I'm confident the change should fix the issue.

Given other examples do this too, if the build passes CI now I don't think it's worth making changes in the short term to get this to run without the macros…but if you're keen, my suspicions sit with that makefile.

So to answer the Q @ltratt, yes — happy with the state of this now — but also happy to make changes if there's anything else to tidy up! I'll squash when instructed.

[1] https://github.com/capablevms/cheri-examples/blob/master/general_bounds.c
[2] https://github.com/capablevms/cheri-examples/commit/ba68b10ed686829e326fb955857718d4cfa53486

0152la:
You're right. It seems it assumes root examples mean `purecap` [1]. Looking at the logic then [2], you should place this example in `./hybrid/`, and remove the additional condition you added in `run_tests.sh`.

I *think* that's the best resolution, but still the fact that we have some tests compiling in `.buildbot.sh`, then others running and compiling in `./tests/run_tests.sh` is not ideal, and should probably fix that. However, that is beyond the scope of this PR. If this works and @ltratt is fine merging as is, without the change at the top, I'm fine with that.

[1] https://github.com/capablevms/cheri-examples/blob/master/.buildbot.sh#L18
[2] https://github.com/capablevms/cheri-examples/blob/master/.buildbot.sh#L29

ltratt:
Yeah, let's get this in, and then refine/move things around in another PR.

@probablytom Please squash.

probablytom:
Squashed!

ltratt:
Please squash.

probablytom:
Squashed.

PR #93, opened 2023-11-28
Try to fix new hang

0152la:
Use `script` to call the CHERI QEMU instances in new subshells, so the `SIGTTY` issue doesn't get propagated to the main shell. Seems to work.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -48,7 +48,9 @@ args=(
     --ssh-port $SSHPORT
     --ssh-key $HOME/.ssh/id_ed25519.pub
 )
-BUILDBOT_PLATFORM=riscv64-purecap python3 tests/run_cheri_examples.py "${args[@]}"
+export BUILDBOT_PLATFORM=riscv64-purecap
+args=${args[@]}
+script -O /dev/null -c "python3 tests/run_cheri_examples.py $args"
 
 echo "Running tests for 'morello-hybrid' using QEMU..."
 args=(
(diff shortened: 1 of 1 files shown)


.buildbot.sh
```diff
@@ -48,7 +48,9 @@ args=(
     --ssh-port $SSHPORT
     --ssh-key $HOME/.ssh/id_ed25519.pub
 )
-BUILDBOT_PLATFORM=riscv64-purecap python3 tests/run_cheri_examples.py "${args[@]}"
+export BUILDBOT_PLATFORM=riscv64-purecap
+args=${args[@]}
+script -c "python3 tests/run_cheri_examples.py $args"
```
ltratt, line 53: This will create a file `typescript` (no, not the modern TypeScript :)). Should we `script -c "python3 tests/run_cheri_examples.py $args" > /dev/null` to avoid that?

0152la: Good point; I had missed that. Fixed in https://github.com/capablevms/cheri-examples/pull/93/commits/d43d04939529b072c7eeec7be1e95017690303d6 (just redirecting was not sufficient from my quick tests).

ltratt:
Assuming this passes your tests, please squash.

0152la:
I've squashed, but the status is not being reported from our CI. It does run the job, but the status response is getting 404'd somewhere. I'm hoping it's some temporary glitch that will iron itself out, but I'll do a bit of testing.

PR #95, opened 2024-04-10
Fix CI not failing

0152la:
Unfortunately, when we moved calling the underlying CHERI instance via `script` (in #93), we omitted to ensure that the return code for the underlying tests was properly returned to the testing script it's being called from. This means we had some underlying silent failures in our CI. We attempt to fix them here, and to ensure future CI checks fail appropriately.

### DIFF
--- a/hybrid/cap_build.c
+++ b/hybrid/cap_build.c
@@ -15,11 +15,12 @@
 //    from the first argument --- which differentiates `cheri_address_set` and
 //   `cheri_cap_build`.
 
-#include "cheriintrin.h"
-#include "include/common.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
+#include "../include/common.h"
+#include "cheriintrin.h"
 
 int main()
 {
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -50,7 +50,7 @@ args=(
 )
 export BUILDBOT_PLATFORM=riscv64-purecap
 args=${args[@]}
-script -O /dev/null -c "python3 tests/run_cheri_examples.py $args"
+script -O /dev/null -ec "python3 tests/run_cheri_examples.py $args"
 
 echo "Running tests for 'morello-hybrid' using QEMU..."
 args=(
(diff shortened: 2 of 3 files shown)


hybrid/cap_build.c
```diff
@@ -15,12 +15,13 @@
...
 
-#include "cheriintrin.h"
-#include "include/common.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
+#include "../include/common.h"
+#include "cheriintrin.h"
+
```
probablytom, line 24: This formatting is different to the rest of the samples in `hybrid/` — perhaps it needs a `clang-format`?

probablytom (commented):
All LGTM, although I wonder whether it needs a `clang-format`?

hybrid/cap_build.c
0152la: I've already applied `clang-format`, and reapplying it doesn't seem to modify the file. If you mean the fact that I split the `include`s, that's just something I do naturally between "system" and "file-specific" includes, and `clang-format` didn't seem to complain. I think there's a switch in `clang-format` to modify this, but I guess we don't turn it on this repo. If you feel strongly about another formatting, I can manually amend it.

probablytom: I don't mind, just assumed it was set by `clang-format` so thought I should point it out. I checked a few examples and nearly all have the imports ordered as this file originally had them, so I'd be tempted to revert that specific change for consistency…

…_but_ I found a couple with slight deviation, so there's no _solid_ convenion. If you'd prefer them ordered this way I'm also happy keeping the change and turning the `clang-format` flag on sometime to make the others consistent with this ordering.

0152la: I don't have strong opinion either way, just a preference for a small a workload as necessary for this PR, but happy to go implement whatever you think is best.

ltratt: You're all being very polite :p I think we should merge as-is, and I'll let @probablytom do the honours!

probablytom: Agree! Let's hit the button 👍️

probablytom (approved):

PR #97, opened 2025-01-16
Fix some failing tests

0152la:
Presumably due to internal CHERI changes, some of our tests were failing CI. They are now updated and fixed.


ltratt:
Thanks!

